PR #1, opened 2018-12-11
TSS-3399 initial files

r-owen:


awclemen:
As Russell has stated, this is just a directory structure for holding the scripts for scriptqueue.  I don't see any problems at all, so I approve the pull request.

PR #2, opened 2019-03-05
Adds atcamera characterization script with flatfield.

tribeiro:


python/lsst/ts/standardscripts/auxtel/detector_characterization/get_std_flat_dataset.py
```diff
@@ -0,0 +1,279 @@
...
+                               f"above maximum allowed {self.maximum_exp_time}s.")
+
+        if filter_id is not None:
+            self.filter = filter_id
+
+        if grating_id is not None:
+            self.grating = grating_id
+
+        if linear_stage is not None:
+            self.linear_stage = linear_stage
```
r-owen: Please consider removing the "if" statements. I think just setting the values should work (the default is None, and that's just fine for these).

python/lsst/ts/standardscripts/auxtel/detector_characterization/get_std_flat_dataset.py
```diff
@@ -0,0 +1,279 @@
...
+        self.read_out_time = 2.
+        self.n_dark = 10
+        self.t_dark = 400.
+        self.n_bias = 10
+        self.n_flat = 2
+        self.flat_base_exptime = 0.5
+        self.flat_dn_range = np.array([1., 2., 4., 8., 16., 32., 64., 128.])
+        self.filter = None
+        self.grating = None
+        self.linear_stage = None
```
r-owen: These lines duplicate the code in `configure`. Consider deleting these and using the values above as the default values for the arguments in `configure` (except the array should be a tuple to avoid mutable arguments). Then check the value provided and, if acceptable, just set the value. In other words:
```
configure(self, n_dark=10, ...):
    if n_dark <= 0:
        raise...
    self.n_dark = n_dark
```
with no need to check for None

python/lsst/ts/standardscripts/auxtel/detector_characterization/get_std_flat_dataset.py
```diff
@@ -0,0 +1,279 @@
...
+
+        if self.grating is not None:
+            await self.checkpoint(f"Configuring ATSpec grating; {self.grating}")
+            self.atspec.cmd_changeDisperser.set(disperser=self.grating)
+            await self.atspec.cmd_changeDisperser.start(timeout=self.cmd_timeout)
+
+        if self.linear_stage is not None:
+            await self.checkpoint(f"Configuring ATSpec linear stage; {self.linear_stage}")
+            self.atspec.cmd_moveLinearStage.set(distanceFromHome=self.linear_stage)
+            await self.atspec.cmd_moveLinearStage.start()
```
r-owen: Please consider a timeout for this command, since waiting without a timeout may leave your script hung forever. If you are worried about really slow moves then please at least add a code comment saying so.

python/lsst/ts/standardscripts/auxtel/detector_characterization/get_std_flat_dataset.py
```diff
@@ -0,0 +1,279 @@
...
+        await self.take_bias_sequence()
+
+        await self.take_flat_sequence()
+
+        await self.take_bias_sequence()
+
+        await self.checkpoint("done")
+
+    async def take_dark_sequence(self):
+        """A coroutine to take the intended sequence of darks.
```
r-owen: Please consider removing description prefixes such as "A coroutine to", "A class to", etc. as it does not add information*. I suggest focusing on conciseness and the active voice, e.g.: "Take the intended sequence of darks."

*one can argue about coroutines because last I checked our sphinx generated docs do not yet label coroutines as such. But it's a known issue that I expect will be fixed.

python/lsst/ts/standardscripts/auxtel/detector_characterization/get_std_flat_dataset.py
```diff
@@ -0,0 +1,279 @@
...
+        """
+        self.log.debug("Taking dark sequence...")
```
r-owen: Consider including the number of darks, e.g. `self.log.debug(f"Taking {self.n_dark} darks")`

Also what do you think of making this an info level message? Naively I would think users would appreciate a message before a series of exposures, but you have the greater experience. And you already have a useful checkpoint.

r-owen: On the other hand...would it make any sense to await a checkpoint either at the start of the loop or for each exposure (in which case the checkpoint name could include the image number)? I have no idea what damage it would do to the calibration to have a pause before a set of flats, say, or between two flats.

python/lsst/ts/standardscripts/auxtel/detector_characterization/get_std_flat_dataset.py
```diff
@@ -0,0 +1,279 @@
...
+
+        timeout = self.read_out_time + self.cmd_timeout
+
+        end_readout_coro = self.atcam.evt_endReadout.next(flush=True,
+                                                          timeout=timeout)
+
+        await self.checkpoint(f"Take image {image_seq_name}")
+
+        await self.atcam.cmd_takeImages.start(timeout=timeout+exp_time)
+        await end_readout_coro
```
r-owen: I'm wondering how likely it is that this will get interleaved with other scripts that are also taking images. If that is a concern then you could make this more robust by comparing the data in the endReadout event.

python/lsst/ts/standardscripts/auxtel/detector_characterization/get_std_flat_dataset.py
```diff
@@ -0,0 +1,279 @@
...
+            A string to identify the image.
+
+        Returns
+        -------
+
+        """
+        self.atcam.cmd_takeImages.set(numImages=1,
+                                      expTime=exp_time,
+                                      shutter=shutter,
+                                      imageSequenceName=str(image_seq_name))
```
r-owen: I am curious why you choose to have take_image take just one image, instead of all the images for a sequence with a single call to takeImages (by setting numImages > 1). Of course what you have will work just fine. But other readers may have the same question, so please consider a code comment explaining.

python/lsst/ts/standardscripts/auxtel/detector_characterization/get_std_flat_dataset.py
```diff
@@ -0,0 +1,279 @@
...
+
+        await self.checkpoint(f"Take image {image_seq_name}")
+
+        await self.atcam.cmd_takeImages.start(timeout=timeout+exp_time)
+        await end_readout_coro
+
+    def set_metadata(self, metadata):
+        metadata.duration = self.n_dark * (
+            self.read_out_time + self.t_dark) + 2. * self.n_bias * self.read_out_time + np.sum(
+            self.n_flat * self.flat_dn_range * (self.read_out_time + self.flat_base_exptime))
```
r-owen: Please consider splitting this up into several different computations. I'm finding it hard to follow.

I had in mind something like this, but there are certainly other ways to do it:
```
dark_time = self.n_dark * ...
bias_time = ...
duration = dark-time + bias_time + 
```

tests/SConscript
```diff
@@ -1,3 +1,10 @@
...
-from lsst.sconsUtils import scripts
+import os
+
+from lsst.sconsUtils import env, scripts
 scripts.BasicSConscript.tests(pyList=[])
+
+for name in ("OSPL_URI", "OPENSPLICE_LOC"):
+    val = os.environ.get(name)
+    if val is not None:
+        env.AppendENVPath(name, val)
```
r-owen: Good catch. I also have this on tickets/DM-17909 so whoever merges gets it.

tests/auxtel/detector_characterization/test_std_flat_dataset.py
```diff
@@ -0,0 +1,81 @@
...
+        self.ndark = 0
+        self.nflat = 0
+
+    async def cmd_take_images_callback(self, id_data):
+        if "bias" in id_data.data.imageSequenceName:
+            self.nbias += 1
+        elif "dark" in id_data.data.imageSequenceName:
+            self.ndark += 1
+        elif "flat" in id_data.data.imageSequenceName:
+            self.nflat += 1
```
r-owen: Really picky but...consider using the same variable names here as in the script. I like `nflat` just fine but it's a bit confusing to have `n_flat` in one place and `nflat`. Either is fine.

tests/auxtel/detector_characterization/test_std_flat_dataset.py
```diff
@@ -0,0 +1,81 @@
...
+        self.at_cam.evt_endReadout.put(self.at_cam.evt_endReadout.DataType())
+
+
+class TestATGetStdFlatDataset(unittest.TestCase):
+
+    def test_script(self):
+        async def doit():
+            harness = Harness()
+
+            # Adds callback to take image command
```
r-owen: Consider removing the comment as unnecessary.

tests/auxtel/detector_characterization/test_std_flat_dataset.py
```diff
@@ -0,0 +1,81 @@
...
+
+            harness.script.set_state(ScriptState.RUNNING)
+
+            harness.script._run_task = asyncio.ensure_future(harness.script.run())
+            await harness.script._run_task
+            harness.script.set_state(ScriptState.ENDING)
+
+            self.assertEqual(harness.nbias, harness.script.n_bias*2)
+            self.assertEqual(harness.ndark, harness.script.n_dark)
+            self.assertEqual(harness.nflat, len(harness.script.flat_dn_range)*harness.script.n_flat)
```
r-owen: What's here is very nice. Please expand it to test your script's calls to ATSpectrograph.

This will be a minor a nuisance as you have to listen to 3 additional commands, and preferably check that they receive the data you expect.

python/lsst/ts/standardscripts/auxtel/detector_characterization/get_std_flat_dataset.py
```diff
@@ -0,0 +1,279 @@
...
+            `flat_dn_range = [1., 2., 4.]`. Default is [1., 2., 4., 8., 16., 32., 64.].
+        filter_id : `int`
+            Id of the filter to use in the ATSpectrograph. If `None` (default) will ignore the setup step.
+        grating_id : `int`
+            Id of the grating to use in the ATSpectrograph. If `None` (default) will ignore the setup step.
+        linear_stage : `float`
+            Position of the linear stage in the ATSpectrograph. If `None` (default) will ignore the
+            setup step.
+        read_out_time :`float`
+            The read out time of the camera, in seconds. Used to estimate script duration.
```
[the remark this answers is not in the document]
tribeiro: Since this is a characterization script, we may be in a situation where the readout time is changed to evaluate the impact on the CCD. So, I think it would be a good idea to leave it as an option to the user to set it up. For most cases, the user is probably ok with leaving it alone though.

python/lsst/ts/standardscripts/auxtel/detector_characterization/get_std_flat_dataset.py
tribeiro: As I understood it, this sequence should not be paused.

tribeiro: I'm actually doing that on `take_image`...

python/lsst/ts/standardscripts/auxtel/detector_characterization/get_std_flat_dataset.py
tribeiro: This is due to an issue with the current version of ATCamera which is not set to take images with `numImages>1`... I added a comment to the code...

PR #4, opened 2019-03-07
DM-17909: Write ATPtg/ATMCS integration test script

r-owen:


r-owen:
To respond to your comment...I don't see any natural place to put additional wait statements. All the steps there should be really quick except when waiting for commands or state, and it yields control whenever it waits. If I am missing a spot please let me know, as I'll be happy to add some sleeps if they are wanted.

PR #7, opened 2019-03-28
DM-18527: Write "slew telescope" script

r-owen:


tribeiro:
When can we close this PR? This code is starting to become stale and we would like to be able to use it.

tribeiro:
@andrewbheyer ?

andrewbheyer:
We left off with this with me trying to run it on the containers on my mac osx, Russell was walking me through the process. We were unsuccessful trying to run `scons`. Correct me if i'm wrong Russell but I believe we concluded that we weren't sure if it was a DDS issue or salobj? 

Perhaps Tiago you can attempt to run the script as well and see what happens? What was bizarre is that everything we tried running was within in a container which was demonstrated to be working by Russell.

PR #10, opened 2019-04-22
DM-19274: Reorganize script in standard and external script repository.

tribeiro:
- Move `calsys_take_narrowband_data.py` script to external scripts.
- Include `python/lsst/ts/standardscripts/auxtel/calsys.py` and `python/lsst/ts/standardscripts/auxtel/latiss.py` modules. These modules implement classes that performs high level operations with a group of components.
- Include unit test for latiss.
- Unit test for calsys need to be implemented
- Modify `get_std_flat_dataset.py` to use the `LATISS` class.
- Add `atcam_take_image` script with unit test.


python/lsst/ts/standardscripts/auxtel/latiss.py
```diff
@@ -0,0 +1,244 @@
...
+            tag = f'bias_{i+1:04}_{nbias:04}'
+
+            if checkpoint is not None:
+                await checkpoint(tag)
+            await self.expose(exp_time=0., shutter=False,
+                              image_seq_name=tag,
+                              science=True, guide=False, wfs=False)
+
+    async def take_darks(self, exptime, ndarks, checkpoint=None):
+        """Take a series of bias images.
```
r-owen: "bias images" -> "dark images"

python/lsst/ts/standardscripts/auxtel/latiss.py
```diff
@@ -0,0 +1,244 @@
...
+            Number of flat frames to take.
+        latiss_filter : `None` or `int` or `str`
+            Filter id or name. If None, ignore setting filter.
+        latiss_grating : `None` or `int` or `str`
+            Grating id or name.  If None, ignore setting grating.
+        latiss_linear_stage : `None` or `float`
+            Linear stage position.  If None, ignore setting linear stage.
+        checkpoint : `coro`
+            A optional coroutine that accepts one string argument that is
+            called before each flat field image is taken.
```
r-owen: A coroutine is what you get when you call `checkpoint(tag)`

I suggest calling it an "awaitable callback", "asynchronous callback" or something like that.

python/lsst/ts/standardscripts/auxtel/latiss.py
```diff
@@ -0,0 +1,244 @@
...
+            await self.take_image(exptime=exptime, shutter=True, image_seq_name=tag,
+                                  latiss_filter=latiss_filter, latiss_grating=latiss_grating,
+                                  latiss_linear_stage=latiss_linear_stage)
+
+    async def take_image(self, exptime, shutter, image_seq_name,
+                         latiss_filter=None, latiss_grating=None, latiss_linear_stage=None,
+                         science=True, guide=False, wfs=False,
+                         ):
+        """Encapsulates the action of setting up the spectrograph and taking
+        images.
```
r-owen: I suggest "Set up the spectrograph and take a series of images."
Say what it does, using the active voice, as you have done for most of these doc strings.

python/lsst/ts/standardscripts/auxtel/latiss.py
```diff
@@ -0,0 +1,244 @@
...
+
+        Parameters
+        ----------
+        exptime : `float`
+            The exposure time for the image, in seconds.
+        shutter : `bool`
+            Should activate the shutter? (False for bias and dark)
+        image_seq_name : `str`
+            A string to identify the image.
+        latiss_filter : `None` or `int` or `str`
```
r-owen: Why the "latiss_" prefix? That sounds like an implementation detail, and if so, I suggest removing it.

python/lsst/ts/standardscripts/auxtel/latiss.py
```diff
@@ -0,0 +1,244 @@
...
+        shutter : `bool`
+            Should activate the shutter? (False for bias and dark)
+        image_seq_name : `str`
+            A string to identify the image.
+        science : `bool`
+            Mark image as science (default=True)?
+        guide : `bool`
+            Mark image as guide (default=False)?
+        wfs : `bool`
+            Mark image as wfs (default=False)?
```
r-owen: Consider using an enum class instead of all these flags. It is simple and it needs to changes to the argument list if another image type is added.

Assuming "science", "guide" and "wfs" are mutually exclusive you can use enum.Enum, if they can mix then enum.Flag.

    class ImageType(enum.Enum):
        SCIENCE = enum.auto()
        GUIDE = enum.auto()
        WAVEFRONT_SENSOR = enum.auto()
        OTHER = enum.auto()  # if this is a valid option

    image_type : `ImageType`
        Image type

then in your code:

    image_type = ImageType(image_type)  # to cast integers to enums and reject invalid values
    if image_type == ImageType.SCIENCE...

Even if they can be used together there are enums class for that

python/lsst/ts/standardscripts/auxtel/latiss.py
```diff
@@ -0,0 +1,244 @@
...
+
+        Returns
+        -------
+        endReadout : `SALPY_ATCamera.ATCamera_logevent_endReadoutC`
+
+        """
+        async with self.cmd_lock:
+            # FIXME: Current version of ATCamera software is not set up to take
+            # images with numImages > 1, so this is fixed at 1 for now and we
+            # loop through any set of images we want to take. (2019/03/11)
```
r-owen: Thank you for the clear note

python/lsst/ts/standardscripts/auxtel/latiss.py
```diff
@@ -0,0 +1,244 @@
...
+            self.atcam.cmd_takeImages.set(numImages=1,
+                                          expTime=float(exp_time),
+                                          shutter=bool(shutter),
+                                          imageSequenceName=str(image_seq_name),
+                                          science=bool(science),
+                                          guide=bool(guide),
+                                          wfs=bool(wfs)
```
r-owen: cmd_takeImages could accept the enum mentioned above, but assuming you'd rather use the XML you have now this could read as:

    science=image_type == ImageType.SCIENCE,
    ....

tests/auxtel/test_cam_take_image.py
```diff
@@ -0,0 +1,62 @@
...
+index_gen = salobj.index_generator()
+
+
+class Harness:
+    def __init__(self):
+        self.index = next(index_gen)
+        self.test_index = next(index_gen)
+        salobj.test_utils.set_random_lsst_dds_domain()
+        self.script = ATCamTakeImage(index=self.index)
+        self.atcam = salobj.Controller(SALPY_ATCamera)
```
r-owen: This is a great start. Some suggestions for a bit more:
- Add controllers for the linear stage, filterwheel, etc., and use callbacks to accumulate information about calls (e.g. a list of requested positions, one per command)
- Accumulate some information in cmd_take_images_callback, such as the requested exposure time, then check it.

Also consider wiring the cmd_take_images_callback to self.atcam in the constructor.

python/lsst/ts/standardscripts/auxtel/detector_characterization/get_std_flat_dataset.py
```diff
@@ -69,23 +70,14 @@ def __init__(self, index):
...
-        self.flat_dn_range = None
-        self.filter = None
-        self.grating = None
-        self.linear_stage = None
-
         # FIXME: Get this parameter from the camera configuration once late
-        # joiner is working
-        # on the open network.
+        # joiner is working on the open network.
         self.maximum_exp_time = 401.  # Maximum exposure time in seconds.
```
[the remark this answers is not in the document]
tribeiro: Only when everyone transition to SAL 3.9. Right now camera is still in 3.8.

python/lsst/ts/standardscripts/auxtel/latiss.py
tribeiro: The problem is that `filter` is a reserved word in python. 

```
filter?
Init signature: filter(self, /, *args, **kwargs)
Docstring:
filter(function or None, iterable) --> filter object

Return an iterator yielding those items of iterable for which function(item)
is true. If function is None, return the items that are true.
Type:           type
Subclasses:
```

python/lsst/ts/standardscripts/auxtel/latiss.py
tribeiro: I'm not entirely sure how these are used in the camera. I'll investigate it thought. I liked your suggestion but I'm not sure I'll be able to implement it in this ticket.

python/lsst/ts/standardscripts/auxtel/latiss.py
tribeiro: That is a very good idea. I'll see what we can come up with the camera folks.

python/lsst/ts/standardscripts/auxtel/latiss.py
r-owen: How about `filterid`?

python/lsst/ts/standardscripts/auxtel/latiss.py
tribeiro: I was talking to the camera folks about this. Actually those flags are to switch on/of the readout of some of the camera CCDs. If wfs=False it will switch off readout of the WFS detectors. Same for guide and science. So, in fact, you can actually switch off the science detector and read the wfs and/or guiding alone...

r-owen: Interesting. So a simple enum isn't a good match. A Flag enum would be fine (since values can be or-d together) or just use the booleans you have, as you prefer.

PR #16, opened 2019-06-06
DM-20045: Fix azimuth comparison in DomeTrajectoryMCS

r-owen:


r-owen:
The code for measuring the wrapping difference between two angles is so simple that I am reluctant to use AstroPy for it.

PR #19, opened 2019-06-11
DM-20089: Update remaining code in ts_standardscripts for scriptqueue 2.1

r-owen:


r-owen:
I think the packages that define the CSCs should handle dependency on `ts_config...` packages, as they are the ones that directly need the configuration packages. I see I am missing those from at least some of the CSCs I wrote, and will use this ticket branch to add them.

PR #28, opened 2019-12-19
DM-22716: update one script to use salobj.current_tai, fix flake8 warnings and update unit tests to use asynctest

r-owen:


python/lsst/ts/standardscripts/auxtel/calsys.py
```diff
@@ -148,7 +148,7 @@ def fiberspectrograph(self):
 
         Returns
         -------
-        large_file_data : ``fiberspectrograph.evt_largeFileObjectAvailable.DataType``
+        large_file_data : ``fiberspectrograph.evt_largeFileObjectAvailable.DataType``  # noqa
```
tribeiro: Is this a typo?

r-owen: The `# noqa` prevents `flake8` from complaining that the line is too long.

It would probably be better to use a different data type but I didn't want to change your text.

PR #34, opened 2020-04-13
DM-24224: Reorganize ts_standardscripts

tribeiro:


python/lsst/ts/standardscripts/auxtel/atcam_take_image.py
```diff
@@ -47,7 +47,7 @@ class ATCamTakeImage(salobj.BaseScript):
     def __init__(self, index):
         super().__init__(index=index, descr="Test ATCamTakeImage")
```
r-owen, line 48: Existing code, but...does this really "Test" anything? If not could you please improve the description? Naively I would expect it to say something like "Take images with ATCamera"

python/lsst/ts/standardscripts/auxtel/enable_atcs.py
```diff
@@ -0,0 +1,79 @@
...
+    def get_schema(cls):
+        schema_yaml = """
+            $schema: http://json-schema.org/draft-07/schema#
+            $id: https://github.com/lsst-ts/ts_standardscripts/auxtel/EnableATTCS.yaml
+            title: EnableLATISS v1
+            description: Enable ATCS. Does not require any configuration.
+            type: object
+            additionalProperties: false
+        """
+        return yaml.safe_load(schema_yaml)
```
r-owen, line 62: The doc string for `BaseScript.get_schema` suggests returning `None` if your script has no configuration. Have found it better to return an empty schema instead, and if so, why? Should I update BaseScript to recommend an empty schema?

If you do decide to change to returning `None` then please look through the other scripts as well for the same boilerplate.

python/lsst/ts/standardscripts/auxtel/integration_tests/dometrajectory_mcs.py
```diff
@@ -487,9 +487,12 @@ async def _track_telescope_loop(self):
                 # TODO DM-24051: ditch this hack
                 # when we no longer need ts_xml 4.8
                 if hasattr(self.atmcs.cmd_trackTarget.DataType(), "taiTime"):
+                    self.log.info("taiTime")
```
r-owen, line 490: Thanks for the TODO. Have we reached this point yet or are we still using XML 4.8?

python/lsst/ts/standardscripts/auxtel/enable_atcs.py
```diff
@@ -0,0 +1,79 @@
...
+
+    async def configure(self, config):
+        """Configure script.
+
+        Parameters
+        ----------
+        config : `types.SimpleNamespace`
+            Script configuration, as defined by `schema`.
+        """
```
r-owen, line 71: This may be a first, but I'm wondering if you really need this doc string, especially since this script has no configuration? In theory sphinx can link to the parent class in salobj, but perhaps we don't have that linking working yet. (if you remove it here you might look at your other new scripts for the same).

python/lsst/ts/standardscripts/auxtel/enable_atcs.py
```diff
@@ -0,0 +1,79 @@
...
+__all__ = ["EnableATTCS"]
+
+import yaml
+
+from lsst.ts import salobj
+from lsst.ts.observatory.control.auxtel.atcs import ATCS, ATCSUsages
+
+
+class EnableATTCS(salobj.BaseScript):
+    """ Enable all ATCS components.
```
r-owen, line 30: Picky point, but please remove the space after the `"""` here and in your other doc strings.

python/lsst/ts/standardscripts/auxtel/slew_telescope_icrs.py
```diff
@@ -82,14 +82,14 @@ def get_schema(cls):
                 type: number
                 minimum: -90
                 maximum: 90
-              rot_pa:
+              rot_sky:
```
r-owen, line 85: Do you want to support `rot_par` as well? If so you could simply omit the default for that. It will either show up as `None` or missing (I'm not sure which) in the config. If it's missing then it'd be easy enough to add it in the `config` method.

(Same question for the weirder one: rot_skyphys or whatever you decide to call it.)

python/lsst/ts/standardscripts/auxtel/startup.py
```diff
@@ -0,0 +1,78 @@
...
+
+__all__ = ["Startup"]
+
+import yaml
+
+from lsst.ts import salobj
+from lsst.ts.observatory.control.auxtel.atcs import ATCS, ATCSUsages
+
+
+class Startup(salobj.BaseScript):
```
r-owen, line 29: Please consider a longer, more descriptive name. (Possibly both here and in ts_observatory_control.

I confess to some surprise that the script (and method in ts_observatory_control) has no configuration. Naively I would have expected you might want to do "everything but open the slit", for example, if you want to exercise the telescope while it is raining or the humidity is too high.

tests/auxtel/test_cam_take_image.py
```diff
@@ -161,8 +161,8 @@ async def test_take_images(self):
         async with self.make_script():
             config = await self.configure_script(
                 nimages=1,
-                exp_times=0,
-                image_type="BIAS",
+                exp_times=1.0,
```
r-owen, line 164: A general comment about these tests. Many of the scripts are not completely trivial *in this package* since all the work is in ts_telescope_control. Would it make any sense to move all these tests into that package and only test the very minimum here, e.g. that the script can be run?

tribeiro: ah yes.. these where pre-existing tests... I was cross to remove them but indeed, it is duplicated since it is using the `LATISS` class...

r-owen: I like tests, but I hate duplication and the maintenance headaches. If the testing is done in the new package I think it would be best not to duplicate complex test code here.

I gently suggest either:
* Making the tests part of the ts_operation_manager library so you can run the same tests here and there without duplicating the test code.
*  Removing tests here that are duplicated in ts_operation_manager (but only when the run method is truly trivial, as it now is for so many scripts).

PR #35, opened 2020-05-08
DM-23801: Fix Contributing entry and modernize the documentation layout.

r-owen:


r-owen:
Looks like breakage from an update to ATCamera_Commands.xml. The `imageType` parameter of the `takeImages` command has been replaced by `keyValuePair`. That is surprising to me; I would expect `imageType` to be fundamental and `keyValuePair` to be a useful addition but not a good replacement. I'll ask around about this.

r-owen:
I will run the CI against ts_xml 5.0 (by creating a temporary tickets/DM-23801 branch of ts_xml) and merge if that passes. @tribeiro will fix the problem with 5.1 on a different ticket.

PR #38, opened 2020-05-15
DM-24912: update one unit test for salobj 5.12; add a test for black formatting; fix .travis.yml

r-owen:


r-owen:
I added black to conda test requires

PR #43, opened 2020-07-10
DM-25824: Rotator/CCW synchronization test

tribeiro:
Add CcwRotatorStressTest.

### DIFF
--- a/python/lsst/ts/standardscripts/maintel/integration_tests/ccw_rotator_stress_test.py
+++ b/python/lsst/ts/standardscripts/maintel/integration_tests/ccw_rotator_stress_test.py
@@ -1,0 +1,266 @@
+# This file is part of ts_standardscripts
+#
+# Developed for the LSST Telescope and Site Systems.
+# This product includes software developed by the LSST Project
+# (https://www.lsst.org).
+# See the COPYRIGHT file at the top-level directory of this distribution
+# for details of code ownership.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+
+__all__ = ["CcwRotatorStressTest"]
+
+import yaml
+import asyncio
+
+import numpy as np
+
+import astropy.units as u
+from astropy.coordinates import Angle
+
+from lsst.ts import salobj
+from lsst.ts.observatory.control.maintel.mtcs import MTCS, MTCSUsages
+
+np.random.seed(14)
+
+
+class CcwRotatorStressTest(salobj.BaseScript):
+    """Perform stress test between Camera Cable Wrap (CCW) and Rotator.
+
+    Parameters
+    ----------
+    index : `int`
+        Index of Script SAL component.
+
+    Notes
+    -----
+    **Checkpoints**
+
+    * rotator_angle: before staring each sequence of random rotator angle.
+
+    * cycle: before each cycle of + and -.
+
+    * reset_rotator: after first test completes, before reseting the Rotator
+        position.
+
+    * delay: Before each "delay test".
+
+    **Details**
+
+    The test consists of:
+
+    * Start by slewing/tracking a circumpolar target with the rotator angle set
+      to be close to zero.
+    * Pick a configurable number of random rotator positions inside the limit.
+    * For each random rotator position P slew to +P, wait for the Rotator to
+      arrive at the position, let Rotator track for 15s and slew to -P. Repeat
+      this 5 times.
+    * Once done, reset to track a target with rotator angle close to zero.
+    * Send a demand to track a target at +80 degrees, but send a new target
+      in the other direction while the rotator is still slewing. Repeat this
+      test after a configurable series of time delay after slewing started.
+
+
+
+    Warning:
+    --------
+    This script requires to slew the telescope. Make sure the system is ready
+    to slew before running the script.
+
+    """
+
+    def __init__(self, index):
+        super().__init__(
+            index=index,
+            descr="Perform stress test between Camera Cable Wrap (CCW) and Rotator.",
+        )
+
+        self.mtcs = MTCS(
+            domain=self.domain, log=self.log, intended_usage=MTCSUsages.Slew
+        )
+
+        # How long to track for before each new slew (in seconds).
+        self.track_time = 15.0
+
+        # Configuration parameters.
+        self.n_rot_pos = 5
+        self.n_pos_repeat = 3
+        self.time_delays = np.array([1.0, 2.0, 5.0, 10.0])
+
+        # Ignore CSCs that are not part of the test. This will allow us to run
+        # the test if these CSCs are not enabled or being used for something
+        # else.
+        self.mtcs.check.mtaos = False
+        self.mtcs.check.mtm1m3 = False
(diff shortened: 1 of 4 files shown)


python/lsst/ts/standardscripts/maintel/integration_tests/ccw_rotator_stress_test.py
```diff
@@ -0,0 +1,266 @@
...
+    Parameters
+    ----------
+    index : `int`
+        Index of Script SAL component.
+
+    Notes
+    -----
+    **Checkpoints**
+
+    * rotator_angle: before staring each sequence of random rotator angle.
```
r-owen, line 49: staring -> starting

Also please consider "angles" instead of "angle".

python/lsst/ts/standardscripts/maintel/integration_tests/ccw_rotator_stress_test.py
```diff
@@ -0,0 +1,266 @@
...
+
+    * cycle: before each cycle of + and -.
```
r-owen, line 51: Please consider "cycle of positive and negative motion"

python/lsst/ts/standardscripts/maintel/integration_tests/ccw_rotator_stress_test.py
```diff
@@ -0,0 +1,266 @@
...
+        position.
+
+    * delay: Before each "delay test".
+
+    **Details**
+
+    The test consists of:
+
+    * Start by slewing/tracking a circumpolar target with the rotator angle set
+      to be close to zero.
```
r-owen, line 63: Why not track in az/el instead? Advantages:
* It makes the test more predictable.
* It simplifies the code since you don't have to compute any kind of rotator offset.
* You can pick any convenient elevation, in case you want to study the effect of gravity. (I suggest making that a configuration parameter with a reasonable default).
* You can slew to the current azimuth, which will potentially speed up the test.

python/lsst/ts/standardscripts/maintel/integration_tests/ccw_rotator_stress_test.py
```diff
@@ -0,0 +1,266 @@
...
+      arrive at the position, let Rotator track for 15s and slew to -P. Repeat
+      this 5 times.
+    * Once done, reset to track a target with rotator angle close to zero.
+    * Send a demand to track a target at +80 degrees, but send a new target
+      in the other direction while the rotator is still slewing. Repeat this
+      test after a configurable series of time delay after slewing started.
+
+
+
+    Warning:
```
r-owen, line 75: `Warning:` -> `Warnings`

to make it a standard section header

python/lsst/ts/standardscripts/maintel/integration_tests/ccw_rotator_stress_test.py
```diff
@@ -0,0 +1,266 @@
...
+    --------
+    This script requires to slew the telescope. Make sure the system is ready
```
r-owen, line 77: Please consider changing "requires to slew" to "slews"

python/lsst/ts/standardscripts/maintel/integration_tests/ccw_rotator_stress_test.py
```diff
@@ -0,0 +1,266 @@
...
+        # Ignore CSCs that are not part of the test. This will allow us to run
+        # the test if these CSCs are not enabled or being used for something
+        # else.
+        self.mtcs.check.mtaos = False
+        self.mtcs.check.mtm1m3 = False
+        self.mtcs.check.mtm2 = False
+        self.mtcs.check.hexapod_1 = False
+        self.mtcs.check.hexapod_2 = False
+        self.mtcs.check.dome = False
+        self.mtcs.check.mtdometrajectory = False
```
r-owen, line 109: Should these be set before `super().__init__`? It might be more obvious that they will have their desired effect if you do that.

python/lsst/ts/standardscripts/maintel/integration_tests/ccw_rotator_stress_test.py
```diff
@@ -0,0 +1,266 @@
...
+            domain=self.domain, log=self.log, intended_usage=MTCSUsages.Slew
+        )
+
+        # How long to track for before each new slew (in seconds).
+        self.track_time = 15.0
+
+        # Configuration parameters.
+        self.n_rot_pos = 5
+        self.n_pos_repeat = 3
+        self.time_delays = np.array([1.0, 2.0, 5.0, 10.0])
```
r-owen, line 98: Please remove this section. These are supplied by the configuration so they should not be set here.

python/lsst/ts/standardscripts/maintel/integration_tests/ccw_rotator_stress_test.py
```diff
@@ -0,0 +1,266 @@
...
+        """Configure the script.
+
+        Parameters
+        ----------
+        config : `types.SimpleNamespace`
+            Script configuration.
+        """
+        self.n_rot_pos = int(config.n_rot_pos)
+        self.n_pos_repeat = int(config.n_pos_repeat)
+        self.time_delays = np.array(config.time_delays)
```
r-owen, line 155: Please consider `self.config = config` to avoid all this duplication.

Also please remove the type casting because your schema specifies the types. If your code actually needed a numpy array then that last cast would be fine, but it only needs a sequence and it is sure to get that.

python/lsst/ts/standardscripts/maintel/integration_tests/ccw_rotator_stress_test.py
```diff
@@ -0,0 +1,266 @@
...
+            index=index,
+            descr="Perform stress test between Camera Cable Wrap (CCW) and Rotator.",
+        )
+
+        self.mtcs = MTCS(
+            domain=self.domain, log=self.log, intended_usage=MTCSUsages.Slew
+        )
+
+        # How long to track for before each new slew (in seconds).
+        self.track_time = 15.0
```
r-owen, line 93: Please make this a configuration parameter (and remove this line as per the next comment).

python/lsst/ts/standardscripts/maintel/integration_tests/ccw_rotator_stress_test.py
```diff
@@ -0,0 +1,266 @@
...
+            time_and_date = await self.mtcs.rem.mtptg.tel_timeAndDate.aget()
+            ra = Angle(time_and_date.lst, unit=u.hourangle)
+            try:
+                await self.mtcs.slew_icrs(
+                    ra=ra,
+                    dec=dec,
+                    rot_sky=-180.0 + 80.0,
+                    stop_before_slew=False,
+                    slew_timeout=delay,
+                )
```
r-owen, line 253: This will stop working in ts_salobj v6, because the slew command will extend the timeout based on the reported INPROGRESS acknowledgement.

Here is one simple fix that will work with salobj 5 and 6 (don't specify a value for the `slew_timeout` argument):
```
await asyncio.wait_for(self.mtcs.slew_icrs(...), timeout=delay)
```

tests/test_maintel_ccw_rotator_stress.py
```diff
@@ -0,0 +1,84 @@
...
+                await self.configure_script(
+                    n_rot_pos=5, n_pos_repeat=3, time_delays=[1.0, 10.0, 100.0]
+                )
+
+    async def test_executable(self):
+        scripts_dir = standardscripts.get_scripts_dir()
+        script_path = (
+            scripts_dir / "maintel" / "integration_tests" / "ccw_rotator_stress.py"
+        )
+        await self.check_executable(script_path)
```
r-owen, line 80: Please add tests that the script actually commands the rotator as expected. I hope the mocks in ts_observatory_control will make that easy, but if not, it's not so bad to create your own mock controllers here or run ts_rotator in simulation mode.

tribeiro:
I'll close this PR. I think you are right that we should be good with a notebook for the time being.

PR #45, opened 2020-08-14
DM-25906: Write some basic Main Telescope Scripts.

tribeiro:
Pass script log attribute to the classes on all Scripts that uses ts_observatory_control classes. This guarantees that the log from the underlying task is logged to the EFD.

Add the following base scripts with generic functionality applicable to both AT/MT and different instruments:

	* base_slew: Abstract slew activities for all TCS classes.
	* base_take_image: Abstract take image activities for all instrument classes (LATISS/ComCam/LSSTCam).
	* enable_group: Abstract enable group of CSCs.
	* standby_group: Abstract standby group of CSCs.

Add the following new scripts for the main telescope:

	* enable_comcam: Enable CSCs that are part of ComCam group.
	* enable_mtcs: Enable CSCs that are part of the MTCS group
	* slew: Perform slew/track.
	* standby_comcam: Put all ComCam CSCs in STANDBY state
	* standby_mtcs: Put all MTCS CSCs in STANDBY state
	* take_image_comcam: Take series of images with ComCam.

Add the following new script for auxiliary telescope:

	* standby_atcs: Put all ATCS CSCs in STANDBY state

Rename the following scripts for auxiliary telescope:

	* slew_telescope_icrs -> slew
	* latiss_take_image -> take_image_latiss


python/lsst/ts/standardscripts/enable_group.py
```diff
@@ -0,0 +1,87 @@
...
+
+    def __init__(self, index, descr):
+        super().__init__(index=index, descr=descr)
+
+        self.config = None
+
+    @property
+    @abc.abstractmethod
+    def group(self):
+        """Return group of CSC attribute.
```
r-owen, line 54: Could you please say more about this? What is a "group of CSC attribute"? What is the data type? I hope it's some existing class you can point to. I want to know what it should contain and how I make one. If it's an existing class I can look at the docs for that class.

python/lsst/ts/standardscripts/enable_group.py
```diff
@@ -0,0 +1,87 @@
...
+        """
+        raise NotImplementedError()
+
+    async def configure(self, config):
```
r-owen, line 58: This class does not return a schema, so please document the rules for that schema here or in the class doc string. In other words: for the parts of the config used by this class, what needs to be in the config?

python/lsst/ts/standardscripts/enable_group.py
```diff
@@ -0,0 +1,87 @@
...
+        self.config = config
+
+    def set_metadata(self, metadata):
+        metadata.duration = 60.0
+
+    async def run(self):
+        if hasattr(self.config, "ignore"):
+            for comp in self.config.ignore:
+                rname = comp.lower().replace(":", "_")
```
r-owen, line 67: I strongly suggest using the same name for ignore and group.components. It should greatly reduce confusion. Any time there is a mismatch you have to document the heck out of it, and folks are likely to still find themselves wondering which version of the name to use where.

I don't think the colon will be a problem for yaml. As long as the colon is embedded in the name, yaml seems to handle it just fine:
```
>>> yaml.safe_load("foo:bar: 1")
{'foo:bar': 1}
```
If you insist on remapping names then please document it here and in the schemas for the subclasses of this class.

python/lsst/ts/standardscripts/maintel/enable_comcam.py
```diff
@@ -0,0 +1,97 @@
...
+            $id: https://github.com/lsst-ts/ts_standardscripts/maintel/enable_mtcs.yaml
+            title: EnableComCam v1
+            description: Configuration for EnableComCam
+            type: object
+            properties:
+                cccamera:
+                    description: Configuration for the CCCamera component.
+                    anyOf:
+                      - type: string
+                      - type: "null"
```
r-owen, line 82: Are the double quotes needed around "null"? If not, I suggest eliminating them, for clarity and consistency.

python/lsst/ts/standardscripts/maintel/enable_mtcs.py
```diff
@@ -0,0 +1,140 @@
...
+    Parameters
+    ----------
+    index : `int`
+        Index of Script SAL component.
+
+    Notes
+    -----
+    **Checkpoints**
+
+    **Details**
```
r-owen, line 57: Please delete this empty section.

python/lsst/ts/standardscripts/maintel/enable_mtcs.py
```diff
@@ -0,0 +1,140 @@
...
+        - MTDomeTrajectory
+
+    Parameters
+    ----------
+    index : `int`
+        Index of Script SAL component.
+
+    Notes
+    -----
+    **Checkpoints**
```
r-owen, line 55: Please either say None or, better yet, add a checkpoint after enabling each CSC.

python/lsst/ts/standardscripts/maintel/slew.py
```diff
@@ -0,0 +1,53 @@
...
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+
+__all__ = ["Slew"]
```
r-owen, line 21: The original name slew_icrs was to indicate that the telescope slewed to a position in ICRS coordinates, as opposed to slewing to a position in el/az. We need both. Please keep ICRS in the name or make this general enough to support both kinds of targets.

tribeiro: So, this script works with ICR coordinates and object name so ICRS would not be a good representation. Maybe I can call this "TrackTarget" instead?

python/lsst/ts/standardscripts/maintel/enable_mtcs.py
tribeiro: All CSCs are enabled concurrently so I can't really put a checkpoint after each one. All say `None`

python/lsst/ts/standardscripts/maintel/enable_comcam.py
tribeiro: Yes, the quotes in `"null"` are required, otherwise `yaml.safe_load` will translate `null` to `None` and the json type should be `null`.

python/lsst/ts/standardscripts/enable_group.py
tribeiro: It should be a subclass of `lsst.ts.observatory.control.base_group.BaseGroup`... I'll document it here...

python/lsst/ts/standardscripts/enable_group.py
tribeiro: good point..

tribeiro:
I think I implemented all the changes... I went ahead and renamed the `Slew` script to be `TrackTarget` script...

python/lsst/ts/standardscripts/enable_group.py
r-owen: Ping

python/lsst/ts/standardscripts/maintel/enable_comcam.py
r-owen: Thank you. That is helpful to know.

python/lsst/ts/standardscripts/maintel/enable_mtcs.py
r-owen: Sounds good.

python/lsst/ts/standardscripts/base_track_target.py
```diff
@@ -0,0 +1,185 @@
...
+
+        Parameters
+        ----------
+        metadata : `Script_logevent_metadata`
+        """
+        metadata.duration = 1
+
+    async def run(self):
+
+        target_name = getattr(self.config, "target_name", "slew_icrs")
```
r-owen, line 144: Do you want to update the default target name, or are you still happy with "slew_icrs"?

python/lsst/ts/standardscripts/enable_group.py
tribeiro: I documented in the class doc....

python/lsst/ts/standardscripts/base_track_target.py
tribeiro: I think `slew_icrs` is a good default... After all, if the user does not provide a target name, it will operate in `slew_icrs` mode.

python/lsst/ts/standardscripts/enable_group.py
r-owen: Ah, found it. Thanks. Yes, that's a good place for it.

python/lsst/ts/standardscripts/base_track_target.py
r-owen: Great. Thanks.

python/lsst/ts/standardscripts/auxtel/enable_atcs.py
```diff
@@ -102,20 +103,11 @@ def get_schema(cls):
                       - type: string
                       - type: "null"
                     default: null
+                ignore:
+                    description: CSCs from the group to ignore.
```
r-owen, line 107: Is this Hexapod:1 or hexapod_1. Please provide an example. I suggest using the former, because I think it is least surprising.

python/lsst/ts/standardscripts/maintel/enable_mtcs.py
r-owen: ping

python/lsst/ts/standardscripts/auxtel/standby_atcs.py
```diff
@@ -0,0 +1,55 @@
...
+    Parameters
+    ----------
+    index : `int`
+        Index of Script SAL component.
+
+    Notes
+    -----
+    **Checkpoints**
+
+    **Details**
```
r-owen, line 39: Please delete this empty section

python/lsst/ts/standardscripts/auxtel/standby_atcs.py
```diff
@@ -0,0 +1,55 @@
...
+    """Put ATCS components in standby.
+
+    Parameters
+    ----------
+    index : `int`
+        Index of Script SAL component.
+
+    Notes
+    -----
+    **Checkpoints**
```
r-owen, line 37: Please add "None" (or similar) as the contents of this section.

Otherwise users may think you forgot to fill in this section.

python/lsst/ts/standardscripts/maintel/enable_comcam.py
```diff
@@ -0,0 +1,97 @@
...
+        - CCArchiver
+
+    Parameters
+    ----------
+    index : `int`
+        Index of Script SAL component.
+
+    Notes
+    -----
+    **Checkpoints**
```
r-owen, line 48: Please add "None"

python/lsst/ts/standardscripts/maintel/enable_comcam.py
```diff
@@ -0,0 +1,97 @@
...
+
+    **Details**
```
r-owen, line 50: Please delete and look for and delete all other empty instances of `**Details**`

tests/test_maintel_enable_mtcs.py
```diff
@@ -0,0 +1,108 @@
...
+                                self.mtcs_mock.controllers, comp
+                            ).evt_summaryState.data.summaryState,
+                            salobj.State.ENABLED,
+                        )
+
+    async def test_ignore_feature(self):
+        async with self.make_script():
+
+            # Test the ignore feature
+
```
r-owen, line 63: Please consider deleting that comment -- the test is called `test_ignore_feature` so I think it adds no new information.

python/lsst/ts/standardscripts/auxtel/enable_atcs.py
```diff
@@ -102,20 +103,13 @@ def get_schema(cls):
                       - type: string
                       - type: "null"
                     default: null
+                ignore:
+                    description: >-
+                        CSCs from the group to ignore. Name must match those in
+                        self.group.components, e.g.; {self.group.components}.
```
r-owen, line 109: Please consider "e.g. Hexapod:1" instead of "e.g.; {self.group.components}." and look for the same text elsewhere. (Even if this does expand to a useful example, I would argue that it is more useful to be able to read the example when reading the source code).

python/lsst/ts/standardscripts/enable_group.py
```diff
@@ -0,0 +1,117 @@
...
+
+
+class EnableGroup(salobj.BaseScript, metaclass=abc.ABCMeta):
+    """Base Script for enabling groups of CSCs.
+
+    This base class is setup to operate with minimum configuration. By default
+    it will try to access an attribute `self.config.ignore`, which is supposed
+    to contain a list of CSCs from the group to be ignored in the process.
+    Nevertheless, if the parent class does not provide an `ignore` property it
+    will be ignored.
```
r-owen, line 35: Please document what the "list of CSCs" will look like. I think the important thing is that each element is a name such as "Hexapod:1" or "hexapod_1".

python/lsst/ts/standardscripts/enable_group.py
```diff
@@ -0,0 +1,117 @@
...
+
+    In addition, when subclassing this base class, users can provide settings
+    for the configurable CSCs in the group. The name of the CSC in the
+    configuration must match the name of the CSC in `group.components`, which
+    is the name of the CSC in lowercase, replacing the ":" with "_" for indexed
+    components. For example,
+
+        * ATMCS -> atmcs
+        * NewMTMount -> newmtmount
+        * Hexapod:1 -> hexapod_1
```
r-owen, line 45: An example of "settings for the configurable CSCs" would be a really big help here.

I think it's just a name with an associated string that is the value for settingsToApply (which I know is going to be renamed). But I'm not sure.

PR #46, opened 2020-08-21
Remove unnecessary ups cfg file, which was causing scons to fail.

tribeiro:


r-owen:
Please fix the Jenkins issue before merging (though I did not look to see what the problem was).

PR #50, opened 2020-09-23
DM-26770: Fix set_summary_state script and add a script to run any command to a CSC.

tribeiro:
Move CSC parsing code in set_summary_state to utils, fix issue with regular expression that cause it to fails with components that had numbers in their name, e.g.; MTM2 and MTM1M3.
Add new script RunCommand, that allows one to run any command from a CSC and (optionally) wait for an event.
Add new option on BaseTrackTarget to “track target” for a specified duration. This allows us to mimic a visit with exposure.


mareuter (approved):
Looks good to me. Nice to have that generic command script.

r-owen:
Wait! Please don't reinvent the CSC name to name, index parser. Let's use the one in salobj. If it needs improving, I am happy to do that.

mareuter:
Didn't know that existed. That sounds like a good idea.

tribeiro:
Good catch! thanks! I'm implementing it.

PR #56, opened 2021-01-14
DM-28187: Update ts_observatory_control and ts_standardscript to Remove NewMTMount.

tribeiro:
Replace NewMTMount with MTMount in enable_mtcs and standby_mtcs scripts.


r-owen (approved):
Excellent. I'm glad it was simple.

PR #60, opened 2021-04-26
tickets/DM-29669

mareuter:


python/lsst/ts/standardscripts/auxtel/offline_latiss.py
```diff
@@ -0,0 +1,69 @@
...
+    def components():
+        """Return list of components name as appeared in
+        `self.group.components`.
+
+        Returns
+        -------
+        components : `list` of `str`.
+
+        """
+        return set(["atcamera", "atspectrograph", "atheaderservice", "atarchiver"])
```
tribeiro, line 69: See my comment about creating a `set` with `{"val1", "val2"}`, above.

python/lsst/ts/standardscripts/maintel/offline_comcam.py
```diff
@@ -0,0 +1,69 @@
...
+    def components():
+        """Return list of components name as appeared in
+        `self.group.components`.
+
+        Returns
+        -------
+        components : `list` of `str`.
+
+        """
+        return set(["cccamera", "ccheaderservice", "ccarchiver"])
```
tribeiro, line 69: See my comment about creating a `set` with `{"val1", "val2"}`, above.

python/lsst/ts/standardscripts/maintel/offline_mtcs.py
```diff
@@ -0,0 +1,80 @@
...
+    def components():
+        """Return list of components name as appeared in
+        `self.group.components`.
+
+        Returns
+        -------
+        components : `list` of `str`.
+
+        """
+        return set(
```
tribeiro, line 67: See my comment about creating a `set` with `{"val1", "val2"}`, above.

tribeiro (approved):
I really liked the way you implemented these scripts. Looks really nice and clean. I made one small comment about how you are creating the sets. Python allows you to create them directly with `{"a", "b"}` instead of converting from a list. Consider implementing it this way. 

Other than that, looks really nice. Thanks for the updates!

PR #77, opened 2022-04-18
Tickets/dm 34116

tribeiro:
Add some scripts to help Scheduler operations. I have derived this set of scripts from our experiencing operating the Scheduler in the last couple runs with the Auxiliary Telescope. The main goal was to make it easier for night crew to interact with the scheduler minimizing the amount of typing required to perform basic tasks and also creating some protections against the most common issues we normally encounter. You will see, for instance, there is a handler for the situation where historical data is lost. This is very particular to some issues we face with DDS from time to time, but it is something that creates a lot of confusion when it happens. So I thought I might add this protection. I know it is probably not the best solution to the problem and I hope we will be able to remove this in the future. Just thought I'd give a foreword about the problem so you have and idea about why it is there.

### DIFF
--- a/python/lsst/ts/standardscripts/auxtel/track_target_and_take_image.py
+++ b/python/lsst/ts/standardscripts/auxtel/track_target_and_take_image.py
@@ -115,19 +115,35 @@ additionalProperties: false
 
         self.tracking_started = True
 
-        await asyncio.gather(
-            self.atcs.slew_icrs(
+        setup_atspec_task = asyncio.create_task(
+            self.latiss.setup_atspec(
+                grating=self.grating[0],
+                filter=f"{self.config.filter_prefix}{self.band_filter[0]}",
+            )
+        )
+
+        try:
+            await self.atcs.slew_icrs(
                 ra=self.config.ra,
                 dec=self.config.dec,
                 rot=self.config.rot_sky,
                 rot_type=RotType.Sky,
                 target_name=self.config.name,
-            ),
-            self.latiss.setup_atspec(
-                grating=self.grating[0],
-                filter=f"{self.config.filter_prefix}{self.band_filter[0]}",
-            ),
-        )
+            )
+        except Exception:
+            self.log.exception(
+                "Failed to slew with required angle. "
+                f"Trying +180 degrees: {180 - self.config.rot_sky}."
+            )
+            await self.atcs.slew_icrs(
+                ra=self.config.ra,
+                dec=self.config.dec,
+                rot=180 - self.config.rot_sky,
+                rot_type=RotType.Sky,
+                target_name=self.config.name,
+            )
+
+        await setup_atspec_task
 
         await self.checkpoint("done")
 
--- a/python/lsst/ts/standardscripts/auxtel/track_target_and_take_image.py
+++ b/python/lsst/ts/standardscripts/auxtel/track_target_and_take_image.py
@@ -160,18 +160,60 @@ additionalProperties: false
     async def _take_data(self):
         """Take data."""
 
-        for exptime, grating, band_filter in zip(
-            self.config.exp_times, self.grating, self.band_filter
-        ):
+        if self.is_standard_visit():
+            self.log.debug(
+                f"Same instrument setup for all images. using n={len(self.config.exp_times)}."
+            )
+            # Note that we are not passing the filter and grating information
+            # take_object, this is handled in track_target_and_setup_instrument
+            # concurrently to slewing to the target. Since all exposures have
+            # the same configuration, we don't need to set them here again.
             await self.latiss.take_object(
-                exptime=exptime,
+                exptime=self.config.exp_times[0],
+                n=len(self.config.exp_times),
                 group_id=self.group_id,
-                filter=f"{self.config.filter_prefix}{band_filter}",
-                grating=grating,
                 reason=self.config.reason,
                 program=self.config.program,
             )
 
+        else:
+            self.log.debug(
+                "Different instrument setup for images. Taking one at a time."
+            )
+
+            for exptime, grating, band_filter in zip(
+                self.config.exp_times, self.grating, self.band_filter
+            ):
+                await self.latiss.take_object(
+                    exptime=exptime,
+                    n=1,
+                    group_id=self.group_id,
+                    filter=f"{self.config.filter_prefix}{band_filter}",
+                    grating=grating,
+                    reason=self.config.reason,
+                    program=self.config.program,
+                )
+
     async def stop_tracking(self):
         """Execute stop_tracking command on ATCS."""
         await self.atcs.stop_tracking()
+
+    def is_standard_visit(self) -> bool:
+        """Determine if script configuration specify a standard visit or a
+        set of observations with different configurations.
+
+        A standard visit consists of a sequence of images with the
+        same instrument configuration and exposure time.
+
+        Returns
+        -------
+        bool
+            True if instrument configuration and exposure time is the same for
+            all exposures.
+        """
+        return all(
+            [
+                len(set(item)) == 1
+                for item in (self.config.exp_times, self.grating, self.band_filter)
+            ]
+        )
--- a/python/lsst/ts/standardscripts/scheduler/set_desired_state.py
+++ b/python/lsst/ts/standardscripts/scheduler/set_desired_state.py
@@ -1,0 +1,223 @@
+# This file is part of ts_standardscripts
+#
+# Developed for the LSST Telescope and Site Systems.
+# This product includes software developed by the LSST Project
+# (https://www.lsst.org).
+# See the COPYRIGHT file at the top-level directory of this distribution
+# for details of code ownership.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+
+__all__ = ["SetDesiredState"]
+
+import types
+import typing
+import asyncio
+
+from lsst.ts import salobj
+from lsst.ts.idl.enums.Scheduler import SalIndex
+
+
+class SetDesiredState(salobj.BaseScript):
+    """A base script that implements setting the desired state for the
+    Scheduler.
+
+    Parameters
+    ----------
+    index : `int`
+        Index of Script SAL component.
+    descr : `str`
+        Script description.
+    scheduler_index : `int`
+        Index of the Scheduler to enable.
+    """
+
+    def __init__(
+        self,
+        index: int,
+        descr: str,
+        scheduler_index: SalIndex,
+        desired_state: salobj.State,
+    ) -> None:
+        super().__init__(index=index, descr=descr)
+
+        self.scheduler_remote = salobj.Remote(
+            domain=self.domain,
+            name="Scheduler",
+            index=scheduler_index,
+            include=["summaryState", "heartbeat"],
+        )
+
+        self.desired_state = desired_state
+
+        self.timeout_start = 30.0
+
+        self.configuration = ""
+
+        self._state_transition_methods_to_try = (
+            self._handle_csc_in_standby,
+            self._handle_csc_in_disabled_or_fault,
+            self._handle_csc_in_enabled,
+        )
+
+    @classmethod
+    def get_schema(cls) -> typing.Optional[typing.Dict[str, typing.Any]]:
+        return None
+
+    async def configure(self, config: types.SimpleNamespace) -> None:
+        pass
+
+    def set_metadata(self, metadata: salobj.type_hints.BaseDdsDataType) -> None:
+        """Set metadata fields in the provided struct, given the
+        current configuration.
+
+        Parameters
+        ----------
+        metadata : ``self.evt_metadata.DataType()``
+            Metadata to update. Set those fields for which
+            you have useful information.
+
+        Notes
+        -----
+        This method is called after `configure` by `do_configure`.
+        The script state will be `ScriptState.UNCONFIGURED`.
+        """
+        metadata.duration = self.timeout_start
+
+    async def assert_liveliness(self) -> None:
+        """Assert that the Scheduler is alive."""
+
+        try:
+            await self.scheduler_remote.evt_heartbeat.next(
+                flush=True,
+                timeout=salobj.base_script.HEARTBEAT_INTERVAL,
+            )
+        except asyncio.TimeoutError:
+            raise AssertionError(
+                f"No heartbeat from Scheduler in the last {salobj.base_script.HEARTBEAT_INTERVAL}s. "
+                "Make sure it is running before trying to enable."
+            )
+
+    def get_summary_state(self) -> typing.Optional[salobj.State]:
+        """Get Scheduler summary state."""
+        current_summary_state = self.scheduler_remote.evt_summaryState.get()
+
+        return (
+            None
+            if current_summary_state is None
+            else salobj.State(current_summary_state.summaryState)
+        )
+
+    async def handle_no_summary_state_data(self) -> None:
+        """Handle condition where no information about Scheduler summary state
+        is available.
+
+        Start by assuming Scheduler is in STANDBY, if this fails, assume it is
+        in DISABLED and finally in ENABLED.
+        """
+
+        for coro in self._state_transition_methods_to_try:
+            if await coro():
+                return
+
+        raise RuntimeError(f"Failed to transition CSC to {self.desired_state!r}.")
+
+    async def _handle_csc_in_standby(self) -> bool:
+        """Handle condition where CSC is in STANDBY."""
+
+        try:
+            await self.scheduler_remote.cmd_start.set_start(
+                configurationOverride=self.configuration,
+                timeout=self.timeout_start,
+            )
+        except salobj.AckError:
+            self.log.warning("Sending start command failed. CSC not in STANDBY")
+            return False
+        else:
+            await self.set_desired_state()
+            return True
+
+    async def _handle_csc_in_disabled_or_fault(self) -> bool:
+        """Handle condition where CSC is either in DISABLED or FAULT."""
+        try:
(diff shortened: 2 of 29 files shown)


python/lsst/ts/standardscripts/scheduler/base_set_desired_state.py
```diff
@@ -0,0 +1,214 @@
...
+        """Enable the Scheduler and exit."""
+
+        await self.checkpoint("Assert liveliness")
+        await self.assert_liveliness()
+
+        current_summary_state = self.get_summary_state()
+
+        if current_summary_state is None:
+            await self.checkpoint("Handling no summary state information")
+            await self.handle_no_summary_state_data()
```
r-owen, line 200: Well you warned me this would be controversial. This seems the wrong place to have such protections. if this is a common problem then perhaps I should add it to salobj set_summary_state.

For now, surely it would suffice to remove all this special handling of "no summary state" and just ask the observers to run the script a second time if it fails the first time?

Adding all this special code here just will make folks wonder why this one works sometimes when all the other scripts that try to set the state of a CSC do not.

python/lsst/ts/standardscripts/scheduler/base_standby.py
```diff
@@ -0,0 +1,47 @@
...
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+__all__ = ["BaseStandBy"]
+
+from lsst.ts import salobj
+
+from .base_set_desired_state import BaseSetDesiredState
+
+
+class BaseStandBy(BaseSetDesiredState):
```
r-owen, line 29: I strongly suggest eliminating all subclasses of BaseSetDesiredState and just implementing the trivial subclasses in the script files.

If you make the desired state a constructor parameter and compute the descr in the base class, that looks like this:
```
from lsst.ts.idl.enums.Scheduler import SalIndex
from lsst.ts.salobj import State
from lsst.ts.standardscripts.scheduler import SetDesiredState

class Standby(SetDesiredState):
    def __init__(self, index):
        super().__init__(index=index, desired_state=State.STANDBY, scheduler_index=SalIndex.AUX_TEL)

asyncio.run(Standby.amain())
```

If I update BaseScript.amain to support kwargs, this simplifies to:
```
from lsst.ts.idl.enums.Scheduler import SalIndex
from lsst.ts.salobj import State
from lsst.ts.standardscripts.scheduler import SetDesiredState

asyncio.run(SetDesiredState.amain(desired_state=State.STANDBY, scheduler_index=SalIndex.AUX_TEL))
```

python/lsst/ts/standardscripts/scheduler/base_set_desired_state.py
```diff
@@ -0,0 +1,214 @@
...
+    Scheduler.
+
+    Parameters
+    ----------
+    index : `int`
+        Index of Script SAL component.
+    descr : `str`
+        Script description.
+    scheduler_index : `int`
+        Index of the Scheduler to enable.
```
r-owen, line 42: I strongly suggest making `desired_state` another constructor argument, just like scheduler_index. That makes the API easier to understand and the value easier to specify (no need to override a method).

python/lsst/ts/standardscripts/scheduler/base_set_desired_state.py
```diff
@@ -0,0 +1,214 @@
...
+    """
+
+    def __init__(self, index: int, descr: str, scheduler_index: int) -> None:
```
r-owen, line 45: I strongly suggest formatting descr in this constructor instead of taking it as an argument for all of these new base classes.

That simplifies use of the class and avoids strings that are wrong (e.g. due to copy/paste errors).

It should be trivial to do as long as you cast the scheduler_index (and desired_state, for this one) into the appropriate enums, so you can use the .name property. That cast has the added advantage of catching invalid values and the somewhat questionable advantage of allowing integer arguments.

python/lsst/ts/standardscripts/scheduler/test_utils/scheduler_controller.py
```diff
@@ -0,0 +1,140 @@
...
+# You should have received a copy of the GNU General Public License
+
+__all__ = ["SchedulerController"]
+
+import asyncio
+
+from lsst.ts import salobj
+
+
+class SchedulerController(salobj.Controller):
```
r-owen, line 28: Please consider class name `MockScheduler`

python/lsst/ts/standardscripts/scheduler/test_utils/scheduler_controller.py
```diff
@@ -0,0 +1,140 @@
+# This file is part of ts_standardscripts
```
r-owen, line 1: I *very strongly* suggest renaming this directory to `testutils`. I've found that pytest tries to run things in any module named `test_foo`, which is a real nuisance.

python/lsst/ts/standardscripts/auxtel/scheduler/enable.py
```diff
@@ -0,0 +1,28 @@
...
+# You should have received a copy of the GNU General Public License
+
+from lsst.ts.standardscripts.scheduler.base_enable import BaseEnable
+
+
+class Enable(BaseEnable):
+    """Enable the ATScheduler."""
+
+    def __init__(self, index: int, descr: str) -> None:
+        super().__init__(index, descr, scheduler_index=2)
```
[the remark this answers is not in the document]
r-owen: These new scheduler-specific subclasses seem too trivial to bother adding to the library. One argument in support of this is the fact that you don't test the subclasses (and I'm not asking you to do so).

I strongly suggest just creating the subclasses in the scripts themselves.

This would work even better if I updated salobj BaseScript so that amain and make_from_cmd_line accepted an extra ``**kwargs`` argument to pass to the base class constructor. I can add that to salobj 7.1 if you like. Then you would not even need trivial subclasses. But for now...

python/lsst/ts/standardscripts/scheduler/base_set_desired_state.py
```diff
@@ -0,0 +1,214 @@
...
+        self.scheduler_remote = salobj.Remote(
+            domain=self.domain,
+            name="Scheduler",
+            index=scheduler_index,
+            include=["summaryState", "heartbeat"],
+        )
+
+        self.timeout_start = 30.0
+
+        self._tasks_to_handle_summary_state = (
```
r-owen, line 57: Please consider reserving "task" for instances of asyncio.Task or Future.

If you keep this code, please consider a name such as `self._state_transition_methods_to_try`

r-owen (changes requested):
This looks very useful.

I am requesting changes because I found one likely bug.

I am also hoping you will be willing to move the new trivial subclasses into the script files and compute the descr arguments in the base classes, but that's up to you.

I won't block on the special "no summary state info" code, but I'm really not happy about it. I'd much rather see a more centralized solution if this is a common problem.

python/lsst/ts/standardscripts/scheduler/base_load_snapshot.py
```diff
@@ -0,0 +1,135 @@
...
+        This method is called after `configure` by `do_configure`.
+        The script state will be `ScriptState.UNCONFIGURED`.
+        """
+        metadata.duration = self.timeout_start
+
+    async def run(self) -> None:
+
+        await self.checkpoint("Loading snapshot")
+        await self.scheduler_remote.cmd_load.set_start(
+            uri=self.snapshot, timeout=self.timeout_start
```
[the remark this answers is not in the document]
tribeiro: So, `uri` is a more generic form or `url`, and is the one the `load` commands supports.

python/lsst/ts/standardscripts/scheduler/base_set_desired_state.py
```diff
@@ -0,0 +1,214 @@
...
+
+        if current_summary_state is None:
+            await self.checkpoint("Handling no summary state information")
+            await self.handle_no_summary_state_data()
+        elif current_summary_state in {salobj.State.ENABLED, salobj.State.DISABLED}:
+            await self.checkpoint("Recycling summary state")
+            self.log.warning(
+                f"Scheduler in {current_summary_state}, recycling summary state."
+            )
+            await self.standby()
```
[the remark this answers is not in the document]
tribeiro: Sorry, why would this fail if the state is ENABLED? I actually test for this condition in the unit test:

https://github.com/lsst-ts/ts_standardscripts/blob/c4f3478fbe64f3ef508f22243927d7bffc6dac5e/tests/test_scheduler_base_enable.py#L198
https://github.com/lsst-ts/ts_standardscripts/blob/c4f3478fbe64f3ef508f22243927d7bffc6dac5e/tests/test_scheduler_base_enable.py#L219

and 

https://github.com/lsst-ts/ts_standardscripts/blob/c4f3478fbe64f3ef508f22243927d7bffc6dac5e/tests/test_scheduler_base_standby.py#L121
https://github.com/lsst-ts/ts_standardscripts/blob/c4f3478fbe64f3ef508f22243927d7bffc6dac5e/tests/test_scheduler_base_standby.py#L142

python/lsst/ts/standardscripts/scheduler/base_set_desired_state.py
tribeiro: I wouldn't classify this as a "common problem" and I don't remember this ever affecting the Scheduler itself.. but, when this error happens users get extremely confused about what it going on and "running the script a second time" won't solve the problem as we still don't have historical data do make it work. 

Believe me, I dislike adding this code as much as you do, probably even more. But, I am trying to make the summit crew night time life easier with these scripts, minimizing the amount of typing required to operate the scheduler and anticipating some issues we've found from time to time. 

I honestly don't know if we should add this to salobj or not. Might be good, might be terrible, I simply don't know... As I said, I never wanted to add it in the first place, but I am fearful enough of this happening during the night that I did..

python/lsst/ts/standardscripts/auxtel/track_target_and_take_image.py
```diff
@@ -144,18 +160,56 @@ async def take_data(self):
...
+            self.log.debug(
+                f"Same instrument setup for all images. using n={len(self.config.exp_times)}."
+            )
             await self.latiss.take_object(
-                exptime=exptime,
+                exptime=self.config.exp_times[0],
+                n=len(self.config.exp_times),
                 group_id=self.group_id,
-                filter=f"{self.config.filter_prefix}{band_filter}",
-                grating=grating,
```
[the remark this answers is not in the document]
tribeiro: right.. this is handled concurrently with the slew command: https://github.com/lsst-ts/ts_standardscripts/blob/c4f3478fbe64f3ef508f22243927d7bffc6dac5e/python/lsst/ts/standardscripts/auxtel/track_target_and_take_image.py#L118-L123

I'll add a note.

python/lsst/ts/standardscripts/auxtel/scheduler/enable.py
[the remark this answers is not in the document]
r-owen: https://jira.lsstcorp.org/browse/DM-34479

python/lsst/ts/standardscripts/scheduler/base_set_desired_state.py
r-owen: If it hasn't affected Scheduler then this seems especially the wrong place for this code.

python/lsst/ts/standardscripts/scheduler/base_set_desired_state.py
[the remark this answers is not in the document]
r-owen: It should will fail because the `standby` command is not accepted in the ENABLED state.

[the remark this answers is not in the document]
tribeiro: the  `standby` method calls `set_summary_state` to send it to standby... https://github.com/lsst-ts/ts_standardscripts/blob/c4f3478fbe64f3ef508f22243927d7bffc6dac5e/python/lsst/ts/standardscripts/scheduler/base_set_desired_state.py#L175-L180

python/lsst/ts/standardscripts/scheduler/base_set_desired_state.py
tribeiro: well, until it happens in the middle of the night... I don't disagree with you.. I will probably remove it.. but it worries me much..

tribeiro: ok, so I started removing this and I was going to add a message to help users work around the problem.. then it hit me, why I added this in the first place.. In order to recover from this problem if the CSC is in STANDBY, they will have to send the start command providing the configuration, which is quite an overhead/copy and pasting around.. more than they would normally do with other CSCs (which now won't require any configuration).. 

The scheduler will always require some configuration for operating for the perceivable future.. so, I think I'll keep this code and hopefully remove it once we either provide a workaround at a lower level or fix this historical data problem (Kafka?)...

r-owen: These two lines will always reconfigure the CSC, if the desired state is DISABLED or ENABLED:

await salobj.set_summary_state(..., state=salobj.State.STANDBY)
await salobj.set_summary_state(..., state=desired_state)

python/lsst/ts/standardscripts/auxtel/track_target_and_take_image.py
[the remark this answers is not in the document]
r-owen: Very nice. Thank you.

python/lsst/ts/standardscripts/scheduler/set_desired_state.py
```diff
@@ -0,0 +1,223 @@
...
+    Scheduler.
+
+    Parameters
+    ----------
+    index : `int`
+        Index of Script SAL component.
+    descr : `str`
+        Script description.
+    scheduler_index : `int`
+        Index of the Scheduler to enable.
```
r-owen, line 42: Thank you for adding the `desired_state` argument. Please document it here.

Also please consider generating the descr in the constructor.

python/lsst/ts/standardscripts/scheduler/testutils/scheduler_controller.py
```diff
@@ -0,0 +1,140 @@
...
+# You should have received a copy of the GNU General Public License
+
+__all__ = ["SchedulerController"]
+
+import asyncio
+
+from lsst.ts import salobj
+
+
+class SchedulerController(salobj.Controller):
```
r-owen, line 28: I strongly encourage renaming this to MockScheduler because the current class name sounds like something that *controls* the scheduler, rather than something that pretends to *be* a Scheduler

scripts/auxtel/scheduler/enable.py
```diff
@@ -0,0 +1,38 @@
...
+#
+# You should have received a copy of the GNU General Public License
+
+import asyncio
+
+from lsst.ts.idl.enums.Scheduler import SalIndex
+from lsst.ts.standardscripts.scheduler.enable import Enable
+
+
+class ATSchedulerEnable(Enable):
```
r-owen, line 28: Thank you for moving this here.

Now a question on timing: I can have a salobj update ready in a day or two that allows kwargs to amain. Would you be interested in using that (and thus further simplifying these new scripts) or is it better to rely on salobj 7.0.0.

python/lsst/ts/standardscripts/scheduler/standby.py
```diff
@@ -0,0 +1,47 @@
...
+
+__all__ = ["Standby"]
+
+from lsst.ts import salobj
+from lsst.ts.idl.enums.Scheduler import SalIndex
+
+from .set_desired_state import SetDesiredState
+
+
+class Standby(SetDesiredState):
```
r-owen, line 30: I suggest ditching this and just subclassing SetDesiredState directly in the new scripts. It's just one more constructor argument and saves two classes that don't seem to be very useful.

python/lsst/ts/standardscripts/scheduler/set_desired_state.py
r-owen: Please consider generating the descr in the constructor instead of taking it as an argument, e.g.:
```
# Cast the inputs to make sure they are valid
scheduler_index = SalIndex(scheduler_index)
desired_state = salobj.State(desired_state)
descr = f"Send {scheduler_index.name} to state {desired_state.name}"
```

scripts/auxtel/scheduler/standby.py
```diff
@@ -0,0 +1,38 @@
...
+#
+# You should have received a copy of the GNU General Public License
+
+import asyncio
+
+from lsst.ts.idl.enums.Scheduler import SalIndex
+from lsst.ts.standardscripts.scheduler.standby import Standby
+
+
+class ATSchedulerStandby(Standby):
```
r-owen, line 28: Please consider subclassing SetDesiredState, instead (and similarly for the other few)

tests/test_auxtel_detector_characterization_std_flat_dataset.py
```diff
@@ -67,14 +69,22 @@ async def cmd_take_images_callback(self, data):
...
+
+            if "bias" in parsed_data["imageType"].lower():
+                self.n_bias += 1
+            elif "dark" in parsed_data["imageType"].lower():
+                self.n_dark += 1
+            elif "flat" in parsed_data["imageType"].lower():
+                self.n_flat += 1
 
-        self.end_readout_tasks.append(asyncio.create_task(self.end_readout()))
+        await self.end_readout_tasks[-1]
```
r-owen, line 87: This suggests that you are sure that the tasks will complete in order. Perhaps consider a comment explaining that.

r-owen (approved):
Looks great. Thanks for reducing the trivial classes.

I have a few minor suggestions I hope you will consider, in addition to the big one of possibly removing the "no summary state" code.

As I state in the comments: if you are willing to use salobj 7.1 I can have it add kwargs to BaseScript.amain, which would simplify your new scripts. TIming is up to you, but I'm happy to push that out right away if it'll do you any good.

python/lsst/ts/standardscripts/scheduler/base_set_desired_state.py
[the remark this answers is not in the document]
r-owen: Got it. I thought it was calling the `standby` command.

python/lsst/ts/standardscripts/scheduler/standby.py
tribeiro: I removed the Standby one but I left the Enabled one. Mostly because the later has configuration in it, which means I would duplicate the code on both auxtel/maintel version and would not be able to unit test it properly.

PR #79, opened 2022-05-06
DM-34130: Add support for additional image type in observatory control package and document them in a more visible place

tribeiro:


mareuter (approved):
Looks good to me.

PR #81, opened 2022-05-16
DM-34817: await SalInfo.start before writing messages

r-owen:


tribeiro (approved):
Looks good! thanks for the updates

PR #85, opened 2022-07-27
Tickets/dm 35737 - Update maintel/SetupMTCS to fix bug when starting m1m3

b1quint:
This is the command being used to raise the M1M3 mirror in `maintel/SetupMTCS`:
```
asyncio.create_task(self.mtcs.raise_m1m3())
```
So it does not wait for the mirror to rise before sending new commands. Need to change it to 
```
await self.mtcs.raise_m1m3()
```

### DIFF
--- a/doc/version_history.rst
+++ b/doc/version_history.rst
@@ -6,6 +6,14 @@
 Version History
 ===============
 
+v1.14.3
+-------
+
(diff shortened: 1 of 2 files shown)


b1quint:
Tested on TTS successfully.

tribeiro (approved):

PR #87, opened 2022-09-21
Tickets/dm 36118

tribeiro:


doc/version_history.rst
```diff
@@ -6,6 +6,26 @@
...
 ===============
 
+v1.15.1
+-------
+
+* In python/lsst/ts/standardscripts/auxtel/track_target_and_take_image.py, implement new abstract method ``check_feasibility``.
+
+* In python/lsst/ts/standardscripts/base_track_target_and_take_image.py, add new ``assert_feasibility`` abstract method to ``BaseTrackTargetAndTakeImage``, that is called before running to verify that the system is in a feasible state to execute the script.
+
+* In python/lsst/ts/standardscripts/maintel/track_target_and_take_image_comcam.py, implement new abstract method ``assert_feasibility``.
```
patrickingraham, line 16: I guess this answers my question in the other PR... but see suggestion below.

python/lsst/ts/standardscripts/auxtel/track_target_and_take_image.py
```diff
@@ -225,3 +225,30 @@ def is_standard_visit(self) -> bool:
                 for item in (self.config.exp_times, self.grating, self.band_filter)
             ]
         )
+
+    async def assert_feasibility(self):
+        """Verify that the system is in a feasible state to execute the
```
patrickingraham, line 230: why not put this in the base class (line 214), then add 3 keyword parameters
TCS=True --> for self.atcs.assert_all_enabled()
LATISS=TRue --> self.latiss.assert_all_enabled()
AOS_LOOPS --> `        ataos_corrections = await self.atcs.rem.ataos.evt_correctionEnabled.aget(
            timeout=self.atcs.fast_timeout
        )`

Then you don't need to copy/paste code all over the place?

python/lsst/ts/standardscripts/auxtel/track_target_and_take_image.py
```diff
@@ -225,3 +225,30 @@ def is_standard_visit(self) -> bool:
...
+        ataos_corrections = await self.atcs.rem.ataos.evt_correctionEnabled.aget(
+            timeout=self.atcs.fast_timeout
+        )
+
+        assert (
+            ataos_corrections.hexapod
+            and ataos_corrections.m1
+            and ataos_corrections.atspectrograph
+        ), (
+            "Required ATAOS corrections not enabled: ",
```
patrickingraham, line 249: same deal...

python/lsst/ts/standardscripts/base_track_target_and_take_image.py
```diff
@@ -205,9 +210,16 @@ async def load_playlist(self):
...
+    async def assert_feasibility(self):
+        """Verify that the system is in a feasible state to execute the
+        script.
+        """
+        raise NotImplementedError()
+
     @abc.abstractstaticmethod
     async def track_target_and_setup_instrument(self):
-        """Implement slewing and setting up instrumment.
+        """Implement slewing and setting up instrument.
```
patrickingraham, line 222: slewing of telescope and setting up of instrument

tests/test_auxtel_track_target_and_take_image.py
```diff
@@ -319,6 +320,25 @@ async def test_run_fail_check_tracking(self):
...
+
+        async with self.make_script(), self.setup_mocks():
+
+            self._ataos_evt_correction_enabled.hexapod = False
+
+            await self.configure_script_full()
+
+            with pytest.raises(AssertionError):
+                await self.run_script()
+
```
patrickingraham, line 333: Maybe add a comment why you make these assertions?
Presumably because you want to make sure nothing moves and the script fails immediately?

patrickingraham (approved):
Please address comments on putting functionality into the base class before merging.
Creation of a 2nd ticket is fine if required.

python/lsst/ts/standardscripts/auxtel/track_target_and_take_image.py
tribeiro: But I agree with you. I am not particularly happy to have the same code duplicated here and there.. I could add `assert_feasibilty` to the `TCS` class but I am on the fence about that.

tests/test_auxtel_track_target_and_take_image.py
tribeiro: Uhm, not sure I understand your question.. Here I am testing that the script will fail the way I want it to fail given the condition imposed by the test. In this case, that it raises an `AssertionError` if the corrections are disabled.

patrickingraham: i meant the assertions below that are not in the `with` statement.
`self.script.atcs.slew_icrs.assert_not_awaited()`

tests/test_auxtel_track_target_and_take_image.py
```diff
@@ -333,9 +353,33 @@ async def setup_mocks(self):
         self.script.latiss.take_object = unittest.mock.AsyncMock(
             side_effect=self.take_object_side_effect
         )
+        self._ataos_evt_correction_enabled = types.SimpleNamespace(
```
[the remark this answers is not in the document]
patrickingraham: How is it local if it's used in line 381?

tests/test_auxtel_track_target_and_take_image.py
tribeiro: ah.. so here I only add to the `with` statement the code that actually belongs there. Because here I want to check that `await self.run_script()` raises `AssertionError` not the rest...

tests/test_auxtel_track_target_and_take_image.py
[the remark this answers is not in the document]
tribeiro: "private" means internal to the class... So methods of the class can call certainly call it, but someone using the class should not. 

It is basically a way of saying, don't use this outside the class.. which might be a bit exaggerated here since this is a test case class, not something people will be reusing..

tests/test_auxtel_track_target_and_take_image.py
patrickingraham: I understand the `with` statement... that's not the issue, sorry for being unclear.
What I mean is that there are 5 assertions below (4 `assert_not_awaited` and 1 `assert_awaited_once` and no explanation of why you are testing those.... and where the "once" comes from.

tests/test_auxtel_track_target_and_take_image.py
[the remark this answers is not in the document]
patrickingraham: ah ok... I thought it meant it should only be used inside the method... not the class.

tests/test_auxtel_track_target_and_take_image.py
tribeiro: Ok, I think I see what you are asking.. The following assertions are basically checking what I expect the code to do when this condition happens.. I though it was a bit self explanatory, but maybe I have that impression because I wrote it.. 

I can add a comment if you feel like it needs some clarification...

PR #90, opened 2022-10-12
DM-36601: make unit tests compatible with salobj 7.2

r-owen:
while retaining backwards compatibility

### DIFF
--- a/doc/version_history.rst
+++ b/doc/version_history.rst
@@ -9,7 +9,9 @@ Version History
 v1.15.2
 -------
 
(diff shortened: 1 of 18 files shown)


doc/version_history.rst
```diff
@@ -6,6 +6,11 @@
 Version History
 ===============
 
+v1.15.3
```
tribeiro, line 9: We never released 1.15.2, so maybe just expand that release?

tribeiro (approved):
Thanks for the updates! looks good!

PR #92, opened 2022-10-14
Tickets/dm 36455

tribeiro:


edennihy (commented):
These changes look good, but should we also add the filter_suffix to the latiss.take_object call for non-standard visits (l203)?

tribeiro:
Good catch! Updated to add suffix when doing non-standard visits..

edennihy (approved):
Thanks, approved!

PR #95, opened 2022-11-03
Update maintel/setup_mtcs.py

b1quint:
* Update maintel/setup_mtcs.py 
  * Now put the mount and the rotator into a disabled state so they can share telemetry.
  * Do the homing of the mount

### DIFF
--- a/python/lsst/ts/standardscripts/maintel/setup_mtcs.py
+++ b/python/lsst/ts/standardscripts/maintel/setup_mtcs.py
@@ -234,8 +234,9 @@ class SetupMTCS(salobj.BaseScript):
             components=["mtmount"],
             overrides=dict(mtmount=self.config.overrides["mtmount"]),
         )
+        # TODO: DM-36932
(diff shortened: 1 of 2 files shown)


python/lsst/ts/standardscripts/maintel/setup_mtcs.py
```diff
@@ -217,6 +234,8 @@ async def start_mtmount(self):
             components=["mtmount"],
             overrides=dict(mtmount=self.config.overrides["mtmount"]),
         )
+        self.log.info("Home mtmount")
+        await self.mtcs.rem.mtmount.cmd_homeBothAxes.set_start()
```
tribeiro, line 238: need to extract this into a method in ts_observatory_control.. Can you add a TODO here and issue a ticket for me to do that?


Also, maybe use `start` instead of `set_start` and add a timeout?

tribeiro (approved):
Left a small comment for you.. please take a look before proceeding..

PR #100, opened 2023-02-06
Update auxtel/daytime_checkout script metadata.duration values

edennihy:


tribeiro (approved):
Looks good. Just, please, update the version history before merging. Use the unreleased v1.18.0.

PR #102, opened 2023-03-03
Tickets/dm 38216

edennihy:


python/lsst/ts/standardscripts/auxtel/daytime_checkout/atpneumatics_checkout.py
```diff
@@ -124,6 +124,9 @@ async def run(self):
 
         await self.atcs.enable_ataos_corrections()
 
+        # Sleep to allow M1 to arrive at commanded pressure
+        await asyncio.sleep(10)
```
tribeiro, line 128: Please, make the sleep time a class variable

python/lsst/ts/standardscripts/auxtel/daytime_checkout/atpneumatics_checkout.py
```diff
@@ -134,6 +137,9 @@ async def run(self):
         # Turn off ATAOS correction(s)
         await self.atcs.disable_ataos_corrections()
 
+        # Sleep to allow M1 to settle down to hard points
+        await asyncio.sleep(15)
```
tribeiro, line 141: As above! Please, make this a variable

python/lsst/ts/standardscripts/auxtel/daytime_checkout/atpneumatics_checkout.py
```diff
@@ -124,22 +138,59 @@ async def run(self):
 
         await self.atcs.enable_ataos_corrections()
 
-        m1_pressure = await self.atcs.rem.atpneumatics.tel_m1AirPressure.aget(timeout=5)
-        self.log.info(
-            f"M1 Air pressure with enabled ATAOS corrections is {m1_pressure.pressure:0.0f} Pascals"
+        # Sleep to allow M1 to arrive at commanded pressure
```
tribeiro, line 141: The way I was thinking this could be implemented is that you enable the correction, then make a loop that waits until the pressure arrives at the commanded pressure (+/- tolerance) and then you give it a delay. 

you may want to turn this into a method though:

```
await self.wait_pressure_buildup()

...

async def wait_pressure_buildup(self):

    time_start = time.time()
    while time.time() < time_start + self.timeout_pressure_buildup:
        (Check pressure, return when arrives in desired window)
    else:
        raise RuntimeError(f"M1 pressure failed to arrive at desired value ({m1_commanded_pressure}+/-{self.pressure_tolerance}) in specified time ({self.timeout_pressure_buildup}s)")
    
```

python/lsst/ts/standardscripts/auxtel/daytime_checkout/atpneumatics_checkout.py
```diff
@@ -124,22 +138,59 @@ async def run(self):
...
         )
 
+        m1_pressure = await self.atcs.rem.atpneumatics.tel_m1AirPressure.aget(
+            timeout=STD_TIMEOUT
+        )
+
+        # Check that M1 pressure arrived to ATAOS commanded pressure within
+        # tolerance
+        if (
+            m1_commanded_pressure.pressure * (1.0 - self.ataos_pressure_tolerance)
```
tribeiro, line 157: you may want to specify that `self.ataos_pressure_tolerance` is a "relative tolerance" I initially thought this was absolute value and got confused. But I agree it makes sense to use relative. So maybe just rename this `self.pressure_relative_tolerance` or `self.pressure_tolerance_relative`?

python/lsst/ts/standardscripts/auxtel/daytime_checkout/atpneumatics_checkout.py
```diff
@@ -124,22 +138,59 @@ async def run(self):
...
+                f"{m1_commanded_pressure.pressure*(1.-self.ataos_pressure_tolerance)} and "
+                f"{m1_commanded_pressure.pressure*(1.+self.ataos_pressure_tolerance)}."
+            )
+
         await self.checkpoint("Turning off ATAOS corrections")
 
         # Turn off ATAOS correction(s)
         await self.atcs.disable_ataos_corrections()
 
+        # Sleep to allow M1 to settle down to hard points
```
tribeiro, line 177: So, similar to the case above, I was thinking we would wait for the pressure to reach the desired value and then give it an additional buffer.

tribeiro (commented):
Some comments for you to consider.. Feel free to pushback on the suggestions.

python/lsst/ts/standardscripts/auxtel/daytime_checkout/atpneumatics_checkout.py
```diff
@@ -79,6 +79,10 @@ def __init__(self, index=1, add_remotes: bool = True):
         self.atcs = ATCS(domain=self.domain, intended_usage=atcs_usage, log=self.log)
         self.main_air_pressure_min_threshold = 275790
         self.main_air_pressure_max_threshold = 413000
+        self.ataos_enabled_delay = 10
```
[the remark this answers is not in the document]
edennihy: Change adopted.

python/lsst/ts/standardscripts/auxtel/daytime_checkout/atpneumatics_checkout.py
edennihy: Change accepted.

python/lsst/ts/standardscripts/auxtel/daytime_checkout/atpneumatics_checkout.py
edennihy: I think the original version is a little simpler so will keep it, but I do like your version of the RuntimeError message.

python/lsst/ts/standardscripts/auxtel/daytime_checkout/atpneumatics_checkout.py
edennihy: Same as above, updated error message to be more informative.

PR #103, opened 2023-03-16
Tickets/dm 38000

edennihy:


tests/test_auxtel_slew_and_take_image_checkout.py
```diff
@@ -42,6 +42,9 @@ def setUp(self) -> None:
         self.ataos_correction_status = True
         self.ataos_correction_result = "enabled"
         self.cmd_time = time.time()
+        self.m3_port_selected = 2
```
patrickingraham, line 45: So the way to do this is not to hard code it but to import the values from here:
https://github.com/lsst-ts/ts_idl/blob/develop/python/lsst/ts/idl/enums/ATMCS.py#L40

Then use variables names (e.g. M3State.Nasmyth2), same goes for the state.

patrickingraham (changes requested):
So the way to do this is not to hard code the enumerations for which numbers correspond to which position, but to import the values from here:
https://github.com/lsst-ts/ts_idl/blob/develop/python/lsst/ts/idl/enums/ATMCS.py#L40

Then use variables names (e.g. M3State.Nasmyth2), same goes for the state.

This will result in fixes throughout....

One annoying thing is that a loss of DDS history might cause problems for this. It'd be good to check that the m3 position is published each time the mount gets enabled.

tests/test_auxtel_slew_and_take_image_checkout.py
edennihy: Ah great, thanks for the pointer. I'll add those changes now.

doc/version_history.rst
```diff
@@ -6,6 +6,13 @@
 Version History
 ===============
 
+v1.19.2
+-------
+
+* In ``auxtel/daytime_checkout/slew_and_take_image_checkout.py``:
+  * add check that M3 is in position for observations with LATISS
+* update unit test ``tests/test_auxtel_slew_and_take_image_checkout.py``
```
patrickingraham, line 14: I think there is an indent problem here

patrickingraham (approved):
Code improved. Thanks. Made one comment.
Might want to consider evaluating the test coverage for this script.

PR #106, opened 2023-04-20
tickets/DM-38797

gmegh:


python/lsst/ts/standardscripts/maintel/laser_tracker/shut_down.py
```diff
@@ -0,0 +1,72 @@
...
+# You should have received a copy of the GNU General Public License
+
+__all__ = ["ShutDown"]
+
+import asyncio
+
+from lsst.ts import salobj
+
+
+class SetUp(salobj.BaseScript):
```
tribeiro, line 28: Rename the Script `SetUp` -> `ShutDown`.

python/lsst/ts/standardscripts/maintel/laser_tracker/shut_down.py
```diff
@@ -0,0 +1,72 @@
...
+        ------
+        asyncio.TimeoutError
+            If the Laser Tracker is not ready after the timeout.
+        """
+        await self.start_up()
+
+        await self.checkpoint("Shutting down Laser Tracker.")
+        await self.scheduler_remote.cmd_powerOff.set_start()
+        await self.checkpoint("Laser Tracker turned off.")
+
```
tribeiro, line 71: I wonder if we should wait for the laser to report as OFF, like we do with the power on... I think it is probably a good idea.

python/lsst/ts/standardscripts/maintel/laser_tracker/shut_down.py
```diff
@@ -0,0 +1,72 @@
...
+        """Run the script.
+        Raises
+        ------
+        asyncio.TimeoutError
+            If the Laser Tracker is not ready after the timeout.
+        """
+        await self.start_up()
+
+        await self.checkpoint("Shutting down Laser Tracker.")
+        await self.scheduler_remote.cmd_powerOff.set_start()
```
tribeiro, line 69: This command will completely power off the laser tracker, later requiring someone to get inside the cell and powering it up again. I think we only want to power off the laser, which is done with the `laserPower` command but passing `power=0`, e.g.;

```
await self.laser_tracker.rem.lasertracker_1.cmd_laserPower.set_start(
             power=0,
             timeout=self.timeout_laser_power,
         )
```

tribeiro (commented):
Please, make sure you reformat the code with black.

python/lsst/ts/standardscripts/maintel/laser_tracker/shut_down.py
gmegh: Done

python/lsst/ts/standardscripts/maintel/laser_tracker/shut_down.py
```diff
@@ -0,0 +1,102 @@
...
+
+    async def configure(self, config):
+        self.config = config
+
+    def set_metadata(self, metadata):
+        """Set estimated duration of the script."""
+
+        metadata.duration = self.timeout_laser_power + self.timeout_laser_warmdow
+
+    async def wait_laser_status_ok(self):
```
gmegh, line 72: should i be calling this method from the set_up and passing OFF as attribute?

tribeiro: You should probably rename the method `wait_laser_status_off`... 

I was thinking about this. We may want to actually have a `LaserTracker` class in ts_observatory_control and put all this functionality there. 

But I don't know.. For now I think we can keep these.

tests/test_maintel_lasertracker_shut_down.py
```diff
@@ -0,0 +1,85 @@
...
+            **{
+                "cmd_laserPower.set_start": self.mock_laser_power,
+                "evt_laserStatus.aget": self.get_laser_status,
+                "evt_laserStatus.next": self.get_laser_status,
+                "evt_summaryState.aget": self.get_summary_state,
+                "evt_summaryState.next": self.get_summary_state,
+            }
+        )
+
+        self.laser_status = types.SimpleNamespace(status=LaserStatus.OFF)
```
tribeiro, line 50: I think you want to start with `LaserStartus.ON`

tribeiro (commented):
Please, remember to update the version history! 

I'm leaving a few inline comments but I think this is pretty close to being done. You can probably rebase/squash the commits into 1 single commit.

python/lsst/ts/standardscripts/maintel/laser_tracker/shut_down.py
gmegh: Done

python/lsst/ts/standardscripts/maintel/laser_tracker/shut_down.py
```diff
@@ -0,0 +1,102 @@
...
+    async def run(self):
+        """Run the script."""
+
+        await self.checkpoint("Shutting down Laser Tracker.")
+        await self.laser_tracker.rem.lasertracker_1.cmd_laserPower.set_start(
+            power=0,
+            timeout=self.timeout_laser_power,
+        )
+
+        await self.checkpoint("Waiting for laser to warm down.")
```
tribeiro, line 99: Consider something like:

```
"Waiting for laser to switch off."
```

It actually does not wait for it to cool down.

tribeiro (commented):
Please, update the version history... I also have one final minor inline comment.

python/lsst/ts/standardscripts/maintel/laser_tracker/shut_down.py
gmegh: Done

PR #107, opened 2023-04-20
Tickets/dm 38564

edennihy:
First attempt at the latiss_take_sequence script. Based on the take_sequence part of the latiss_acquire_and_take_sequence script in ts_externalscripts. The other part (latiss_acquire) will live in ts_externalscripts.

### DIFF
--- a/doc/version_history.rst
+++ b/doc/version_history.rst
@@ -6,6 +6,11 @@
 Version History
 ===============
 
+v1.20.0
+-------
+
+* Add new ``auxtel/latiss_take_sequence.py`` script, unit tests, and executables
+
 v1.19.2
 -------
 
--- a/python/lsst/ts/standardscripts/auxtel/latiss_take_sequence.py
+++ b/python/lsst/ts/standardscripts/auxtel/latiss_take_sequence.py
@@ -1,0 +1,251 @@
+# This file is part of ts_standardscripts
+#
+# Developed for the LSST Telescope and Site Systems.
+# This product includes software developed by the LSST Project
+# (https://www.lsst.org).
+# See the COPYRIGHT file at the top-level directory of this distribution
+# for details of code ownership.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+__all__ = ["LatissTakeSequence"]
+
+import asyncio
+
+import collections
+
+import yaml
+from lsst.ts import salobj
+from lsst.ts.observatory.control.auxtel import ATCS, LATISS, ATCSUsages, LATISSUsages
+from lsst.ts.standardscripts.utils import format_as_list
+
+from lsst.ts.idl.enums.Script import ScriptState
+
+STD_TIMEOUT = 20  # seconds
+
+
+class LatissTakeSequence(salobj.BaseScript):
+    """Perform a sequence of exposures for a given set of instrument
+    configurations.
+
+    Parameters
+    ----------
+    index : `int`
+        Index of Script SAL component.
+
+    Notes
+    -----
+    An (optional) checkpoint is available to verify the calculated
+    telescope offset after each iteration of the acquisition.
+
+    **Details**
+
+    """
+
+    def __init__(self, index, add_remotes: bool = True):
+        super().__init__(
+            index=index,
+            descr="Perform a sequence of images with LATISS.",
+        )
+
+        latiss_usage = None if add_remotes else LATISSUsages.DryTest
+
+        atcs_usage = None if add_remotes else ATCSUsages.DryTest
+
+        self.latiss = LATISS(
+            domain=self.domain, intended_usage=latiss_usage, log=self.log
+        )
+        self.atcs = ATCS(domain=self.domain, intended_usage=atcs_usage, log=self.log)
+
+        self.run_started = False
+
+    @classmethod
+    def get_schema(cls):
+        schema_yaml = """
+            $schema: http://json-schema.org/draft-07/schema#
+            $id: https://github.com/lsst-ts/ts_externalscripts/auxtel/latiss_take_sequence.yaml
+            title: LatissAcquireAndTakeSequence v1
+            description: Configuration for LatissAcquireAndTakeSequence Script.
+            type: object
+            properties:
+              program:
+                description: Name of the program this dataset belongs to (required).
+                type: string
+
+              reason:
+                description: Reason for taking the data (required).
+                type: string
+
+              filter_sequence:
+                description: Filters for exposure sequence. If a single value is specified then
+                   the same filter is used for each exposure.
+                anyOf:
+                  - type: array
+                    minItems: 1
+                    items:
+                      type: string
+                  - type: string
+                default: empty_1
+
+              grating_sequence:
+                description: Gratings for exposure sequence. If a single value is specified then
+                   the same grating is used for each exposure.
+                anyOf:
+                  - type: array
+                    minItems: 1
+                    items:
+                      type: string
+                  - type: string
+                default: empty_1
+
+              exposure_time_sequence:
+                description: Exposure times for exposure sequence (sec). Each exposure requires
+                   a specified exposure time.
+                anyOf:
+                  - type: array
+                    minItems: 1
+                    items:
+                      type: number
+                      minimum: 0
+                  - type: number
+                    minimum: 0
+                default: 2.
+
+              do_check_ataos_corrections:
+                description: Check that ATAOS corrections are enabled before taking sequence.
+                type: boolean
+                default: True
+            required: ["program", "reason"]
+        """
+        return yaml.safe_load(schema_yaml)
+
+    async def configure(self, config):
+        """Configure script.
+
+        Parameters
+        ----------
+        config : `types.SimpleNamespace`
+            Script configuration, as defined by `schema`.
+        """
+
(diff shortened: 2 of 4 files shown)


python/lsst/ts/standardscripts/auxtel/latiss_take_sequence.py
```diff
@@ -0,0 +1,295 @@
...
+from lsst.ts.standardscripts.utils import format_as_list
+
+from lsst.ts.idl.enums.Script import ScriptState
+
+STD_TIMEOUT = 20  # seconds
+
+
+class LatissTakeSequence(salobj.BaseScript):
+    """
+    Perform a sequence of exposures for a given
```
tribeiro, line 40: Please, move this phrase to the line above...

python/lsst/ts/standardscripts/auxtel/latiss_take_sequence.py
```diff
@@ -0,0 +1,295 @@
...
+        super().__init__(
+            index=index,
+            descr="Perform a sequence of images with LATISS.",
+        )
+
+        latiss_usage = None if add_remotes else LATISSUsages.DryTest
+
+        atcs_usage = None if add_remotes else ATCSUsages.DryTest
+
+        # Instantiate latiss. We need to do this after the call to
```
tribeiro, line 67: I think you can remove this comment.

python/lsst/ts/standardscripts/auxtel/latiss_take_sequence.py
```diff
@@ -0,0 +1,295 @@
...
+                format_as_list(config.filter_sequence, _recurrences),
+                format_as_list(config.exposure_time_sequence, _recurrences),
+                format_as_list(config.grating_sequence, _recurrences),
+            )
+        ]
+
+        self.reason = config.reason
+
+        if self.reason is None:
+            warnings.warn(
```
tribeiro, line 177: Uhm.. I think creating a new script with a deprecated feature is probably not a good idea. Just enforce it from the get go.

python/lsst/ts/standardscripts/auxtel/latiss_take_sequence.py
```diff
@@ -0,0 +1,295 @@
...
+                "Reason not specified! Consider specifying the 'reason' for "
+                "this observation. it helps reducing the data aftwards. Running "
+                "without specifying a 'reason' will stop working in the future!",
+                DeprecationWarning,
+            )
+
+        self.program = config.program
+
+        if self.program is None:
+            warnings.warn(
```
tribeiro, line 187: again, new script should not be created with deprecated features...

python/lsst/ts/standardscripts/auxtel/latiss_take_sequence.py
```diff
@@ -0,0 +1,295 @@
...
+                "Program not specified! Consider specifying the 'program' for "
+                "this observation, it helps reducing the data aftwards. Running "
+                "without specifying a 'program' will stop working in the future!",
+                DeprecationWarning,
+            )
+
+        self.do_track_target = config.do_track_target
```
tribeiro, line 194: I thought we agreed that this could be removed.. this script is designed to take data while tracking. So, this should not be optional.

python/lsst/ts/standardscripts/auxtel/latiss_take_sequence.py
```diff
@@ -0,0 +1,295 @@
...
+
+              program:
+                description: >-
+                  Optional name of the program this dataset belongs to.
+                anyOf:
+                  - type: string
+                  - type: "null"
+                default: null
+
+              do_track_target:
```
tribeiro, line 137: I think you should remove this option. This Script should only run if we are tracking.

python/lsst/ts/standardscripts/auxtel/latiss_take_sequence.py
```diff
@@ -0,0 +1,295 @@
...
+            filters.add(filt)
+            gratings.add(grating)
+            metadata.duration += 3  # time to reconfigure latiss
+        metadata.duration += exptime_total
+        metadata.filter = f"{filters},{gratings}"
+
+    async def take_sequence(self):
+        """Take data while making sure ATCS is tracking if do_track_target."""
+
+        if self.do_track_target:
```
tribeiro, line 212: I think we should remove this conditional here. Then, before attempting to take data, we should probably check that the telescope is tracking. 

In short:

1. check if we are tracking
2. take sequence and make sure we continue to track.

python/lsst/ts/standardscripts/auxtel/latiss_take_sequence.py
```diff
@@ -0,0 +1,295 @@
...
+                await self.atcs.stop_tracking()
+            except asyncio.TimeoutError:
+                self.log.exception(
+                    "Stop tracking command timed out during cleanup procedure."
+                )
+            except Exception:
+                self.log.exception("Unexpected exception in stop_tracking.")
+
+            try:
+                await self.atcs.disable_ataos_corrections()
```
tribeiro, line 293: You should not do this here.

tribeiro (commented):
Initial round of comments.

python/lsst/ts/standardscripts/auxtel/latiss_take_sequence.py
edennihy: Conditional removed as part of latest commits, is the tasks grouping (L177) enough to do the check before and check while, or are you recommending I add another, separate check?

python/lsst/ts/standardscripts/auxtel/latiss_take_sequence.py
```diff
@@ -0,0 +1,248 @@
...
+    async def assert_feasibility(self) -> None:
+        """Verify that the telescope and camera are in a feasible state to
+        execute the script.
+        """
+
+        await self.latiss.assert_all_enabled()
+        await self.atcs.assert_all_enabled()
+
+        if self.do_check_ataos_corrections:
+            self.log.debug("Check ATAOS corrections are enabled.")
```
tribeiro, line 214: I made a comment on the other PR about this.. we should move this to the `ATCS` class.. if you don't want to do it now, please, add a TODO and file a ticket.

python/lsst/ts/standardscripts/auxtel/latiss_take_sequence.py
tribeiro: I think this is good for now.. But, I think we will want to make it a bit better in the future, adding a `self.atcs.assert_tracking()`, which currently does not exists.

Can you create a ticket to "Create a method in `BaseTCS` to assert that the telescope is tracking."? Then add a TODO here:

```
# TODO (DM-12345): Call self.atcs.assert_tracking().
```

tests/test_auxtel_latiss_take_sequence.py
```diff
@@ -0,0 +1,176 @@
...
+        yield
+
+    async def test_executable(self):
+        scripts_dir = get_scripts_dir()
+        script_path = scripts_dir / "auxtel" / "latiss_take_sequence.py"
+        logger.debug(f"Checking for script in {script_path}")
+        await self.check_executable(script_path)
+
+    async def test_configure(self):
+        async with self.make_script(), self.setup_mocks():
```
tribeiro, line 88: Can you separate the different tests into separate test methods with names that explain what you are testing?

Basically, put every `async with self.make_script(), self.setup_mocks():`  in a different test function.

doc/version_history.rst
```diff
@@ -6,6 +6,11 @@
 Version History
 ===============
 
+v1.19.3
```
tribeiro, line 9: I think adding a new script should probably be a new minor release.. 

we have several PRs open so we should make this batch `v1.20.0`. So, once you merge this PR we should wait.

tests/test_auxtel_latiss_take_sequence.py
```diff
@@ -0,0 +1,176 @@
...
+import unittest
+import logging
+
+import pytest
+from lsst.ts import salobj
+from lsst.ts.standardscripts.auxtel import LatissTakeSequence
+from lsst.ts.standardscripts import BaseScriptTestCase, get_scripts_dir
+
+
+logger = logging.getLogger(__name__)
```
tribeiro, line 32: Instead of doing this, add the logger to the class.. 

```
    @classmethod
    def setUpClass(cls) -> None:
        cls.log = logging.getLogger(type(self).__name__)
```

Then replace all calls to `logger` to `self.log`...

tests/test_auxtel_latiss_take_sequence.py
```diff
@@ -0,0 +1,176 @@
...
+        )
+
+    @contextlib.asynccontextmanager
+    async def setup_mocks(self):
+        self.script.latiss.take_object = unittest.mock.AsyncMock()
+        self.script.atcs.check_tracking = unittest.mock.AsyncMock()
+        self.script.atcs.stop_tracking = unittest.mock.AsyncMock()
+        self.script.atcs.disable_ataos_corrections = unittest.mock.AsyncMock()
+
+        self.script.atcs.rem = types.SimpleNamespace(ataos=unittest.mock.AsyncMock())
```
tribeiro, line 73: Consider doing:

```
self.script.atcs.rem.ataos = unittest.mock.AsyncMock()
```

instead

python/lsst/ts/standardscripts/auxtel/latiss_take_sequence.py
edennihy: Ticket created, see https://jira.lsstcorp.org/browse/DM-38822. After creating the ticket, I also found https://jira.lsstcorp.org/browse/DM-24488, which I think encompasses this method (and maybe some more?). I linked the two tickets in JIRA, but we may want to deprecate one of them.

python/lsst/ts/standardscripts/auxtel/latiss_take_sequence.py
edennihy: TODO added and ticket generated https://jira.lsstcorp.org/browse/DM-38823

tribeiro (approved):
Alright! Looks good! A couple things before you merge:

1. Please, squash the commits.
2. Add the following line to all the license headers: 
    ```
    # along with this program.  If not, see <https://www.gnu.org/licenses/>.
   ```

    This is missing on all your new files...

PR #109, opened 2023-05-02
Tickets/dm 38930

edennihy:


doc/version_history.rst
```diff
@@ -6,10 +6,17 @@
 Version History
 ===============
 
+v1.21.0
```
tribeiro, line 9: We have not released 1.20.0 yet, so you can make these new scripts part of that release instead.

python/lsst/ts/standardscripts/base_offset_tcs.py
```diff
@@ -0,0 +1,200 @@
...
+        self.reset_offsets = getattr(config, "reset_offsets", None)
+
+        self.relative = config.relative
+        self.absorb = config.absorb
+
+    def set_metadata(self, metadata):
+        metadata.duration = 10
+
+    async def assert_feasibility(self) -> None:
+        """Verify that the telescope and is in a feasible state to
```
tribeiro, line 158: remove the "and" after "telescope"?


> Verify that the telescope ~and~ is in a feasible...

python/lsst/ts/standardscripts/base_offset_tcs.py
```diff
@@ -0,0 +1,200 @@
...
+        execute the script.
+        """
+
+        await self.tcs.assert_all_enabled()
+
+    async def run(self):
+        await self.assert_feasibility()
+
+        if self.offset_azel:
```
tribeiro, line 167: I think it is better if you check that this is not None. Doing it this way may give the impression that `self.offset_azel` is a boolean. Also, I think it is not safe to assume that `None==False`.

Consider changing these to:

```
if self.offset_azel is not None:
    ...
```

tribeiro (commented):
Here some initial comments. I want to double check this configuration before approving the PR.

tests/test_maintel_offset_mtcs.py
```diff
@@ -0,0 +1,187 @@
...
+    async def test_offset_rot(self):
+        async with self.make_script():
+            config = dict(offset_rot=dict(rot=10))
+
+            await self.configure_script(**config)
+
+            self.script.mtcs.offset_rot = unittest.mock.AsyncMock()
+
+            await self.script.run()
+
```
tribeiro, line 165: Make sure you assert that `self.script.mtcs.offset_rot` was awaited with the expected value.

tribeiro (approved):
Alright! I verified the configuration schema is actually doing what we need it to do, so all good!

One additional comments for you to consider before merging.

doc/version_history.rst
edennihy: Updated to v1.20.1

tests/test_maintel_offset_mtcs.py
edennihy: I think that's being done in L166 below, right?

PR #115, opened 2023-05-18
Tickets/dm 39179

edennihy:


gmegh:
Is this PR substituting in a way https://github.com/lsst-ts/ts_externalscripts/blob/develop/python/lsst/ts/externalscripts/auxtel/latiss_intra_extra_focal_data.py? 
Or will it require any changes to the mentioned ts_externalscript?

python/lsst/ts/standardscripts/auxtel/offset_ataos.py
```diff
@@ -0,0 +1,155 @@
...
+    def set_metadata(self, metadata):
+        metadata.duration = 10
+
+    async def assert_feasibility(self) -> None:
+        """Verify that the telescope is in a feasible state to
+        execute the script.
+        """
+
+        await self.atcs.assert_all_enabled()
+
```
tribeiro, line 143: We should also assert that the ataos corrections are enabled. 

In the track_target script we do [this](https://github.com/lsst-ts/ts_standardscripts/blob/05ad34eee66b7e7b3f29bf6ea5b301cdd9ce0e36/python/lsst/ts/standardscripts/auxtel/track_target_and_take_image.py#L236-L251), but we should probably move this to `ATCS` and call it here.

python/lsst/ts/standardscripts/auxtel/offset_ataos.py
```diff
@@ -0,0 +1,155 @@
...
+              reset_offsets:
+                type: array
+                description: >-
+                    List of axes to reset. Must be one of z, x, y, z, u, v,
+                    m1, all
+                items:
+                    type: string
+
+            additionalProperties: false
+            anyOf:
```
tribeiro, line 95: Why only allow one offset at a time? I can see a reason to either allow offsetting or resetting, but one should be able to do more than one offset at the same time.

python/lsst/ts/standardscripts/auxtel/offset_ataos.py
```diff
@@ -0,0 +1,155 @@
...
+    async def configure(self, config):
+        """Configure script.
+
+        Parameters
+        ----------
+        config : `types.SimpleNamespace`
+            Script configuration, as defined by `schema`.
+        """
+
+        self.offset_z = getattr(config, "z", 0.0)
```
tribeiro, line 115: You can probably skip these assignments and go directly to the `self.offsets` dictionary in line 125. I suggest doing something like this:

```
self.offset = dict(**[getattr(config, axis, 0.0) for axis in ["x", "y", "z", "u", "v", "m1"])
```

python/lsst/ts/standardscripts/auxtel/offset_ataos.py
```diff
@@ -0,0 +1,155 @@
...
+                description: When correcting coma, also offset the telescope?
+                type: boolean
+                default: true
+              reset_offsets:
+                type: array
+                description: >-
+                    List of axes to reset. Must be one of z, x, y, z, u, v,
+                    m1, all
+                items:
+                    type: string
```
tribeiro, line 92: This should probably be an enum since you have a fixed set of valid values

tribeiro (commented):
Initial round of comments.. this looks really good.

python/lsst/ts/standardscripts/auxtel/offset_ataos.py
edennihy: I think the anyOf call here is saying we require at a minimum one of the following parameters be set. So you could still do an x,y,z offset in one. This case is actually covered in the unit test.

edennihy:
@gmegh, no this does not substitute or replace the latiss_intra_extra_focal_data script. It is for a separate use case, so no changes needed to the other script.

python/lsst/ts/standardscripts/auxtel/offset_ataos.py
edennihy: I will add the standard block of commands here and the new ATCS method in a separate ticket https://jira.lsstcorp.org/browse/DM-39297

doc/version_history.rst
```diff
@@ -7,10 +7,10 @@ Version History
 ===============
 
 v1.21.0
-_______
+-------
```
tribeiro, line 10: thanks for fixing this..

python/lsst/ts/standardscripts/auxtel/offset_ataos.py
tribeiro: ah! `anyOf`, for some reason I thought it was `oneOf`, my bad.

python/lsst/ts/standardscripts/auxtel/offset_ataos.py
tribeiro: Please, add a "TODO" comment here. No need to update the other instances where this occurs, but let's have it at least here, e.g.;

```
# TODO (DM-39297): Replace this with call to assert ataos corrections are enabled in ATCS.
```

python/lsst/ts/standardscripts/auxtel/offset_ataos.py
```diff
@@ -0,0 +1,160 @@
...
+            "The following loops must all be closed (True), but are currently: "
+            f"Hexapod: {ataos_corrections.hexapod}, "
+            f"M1: {ataos_corrections.m1}, "
+            f"ATSpectrograph: {ataos_corrections.atspectrograph}. "
+            "Enable corrections with the ATAOS 'enableCorrection' command before proceeding.",
+        )
+
+    async def run(self):
+        await self.assert_feasibility()
+
```
tribeiro, line 151: Can you add a couple checkpoints please?

tests/test_auxtel_offset_ataos.py
```diff
@@ -0,0 +1,152 @@
...
+
+    async def basic_make_script(self, index):
+        self.script = OffsetATAOS(index=index, add_remotes=False)
+
+        return (self.script,)
+
+    @contextlib.asynccontextmanager
+    async def setup_mocks(self):
+        self.script.atcs.offset_aos_lut = unittest.mock.AsyncMock()
+        self.script.atcs.rem = types.SimpleNamespace(ataos=unittest.mock.AsyncMock())
```
tribeiro, line 47: I think instead of replacing `self.script.atcs.rem` entirely with the namespace mock, you can probably just override `self.script.atcs.rem.ataos`,  e.g.:

```
self.script.atcs.rem.ataos = unittest.mock.AsyncMock()
```

python/lsst/ts/standardscripts/auxtel/offset_ataos.py
```diff
@@ -0,0 +1,160 @@
...
+                type: number
+                description: M1 pressure offset (Pa).
+                maximum: 0
+              offset_telescope:
+                type: bool
+                description: When correcting coma, also offset the telescope?
+                type: boolean
+                default: true
+              reset_offsets:
+                type: array
```
tribeiro, line 87: Ok, slightly picky comment but I think the "all" option could be done directly as a string/enum instead of part of the array/string/enum. So something like this:

```
oneOf:
    -
        type: array
        minItems: 1
        uniqueItems: true
        items:
            type: string
            enum: ["z", "x", "y", "u", "v", "m1"]
    -
        type: string
        enum: ["all"]
``` 


Then, the config will either be:

```
reset_offsets: all
```

or

```
reset_offsets: ["x", "y", "z"]
```

and you can check that things like this:

```
reset_offsets: ["x", "x"]
```

would fail.

tribeiro (commented):
a couple more comments..

tribeiro (approved):
Looks good! thanks for the updates!

PR #116, opened 2023-05-19
DM-39206 - Power on ATCalSys

isotuela:


python/lsst/ts/standardscripts/auxtel/calibrations/power_on_atcalsys.py
```diff
@@ -0,0 +1,330 @@
...
+        # # Confirm lamp turns on (It will go into a warm up period before it will turn on)
+        # try:
+        #     await asyncio.wait_for(
+        #         self.has_lamp_finish_warm_up(), self.lamp_warm_up_time
+        #     )
+        # except asyncio.TimeoutError:
+        #     self.log.info(f"Lamp didn't turn on after {self.lamp_warm_up_time} s")
+
+    async def wait_for_lamp_to_warm_up(self):
+        await self.white_light_source.evt_lampState.flush()
```
tribeiro, line 263: `flush` is a regular method not a coroutine. You should drop the `await`.

python/lsst/ts/standardscripts/auxtel/calibrations/power_on_atcalsys.py
```diff
@@ -0,0 +1,330 @@
...
+        while lamp_state.basicState != ATWhiteLight.LampBasicState.ON:
+            lamp_state = await self.white_light_source.evt_lampState.next(
+                flush=False,
+                timeout=self.timeout_lamp_warm_up
+            )
+            if lamp_state.basicState == ATWhiteLight.LampBasicState.WARMUP:
+                warmup_time_left = (
+                    lamp_state.warmupEndTime - lamp_state.private_rcvStamp
+                )
+                self.log.info(
```
tribeiro, line 278: Maybe use f-string?

```
f"Time Left for lamp warmup: {warmup_time_left} min."
```

python/lsst/ts/standardscripts/auxtel/calibrations/power_on_atcalsys.py
```diff
@@ -0,0 +1,330 @@
...
+        # lamp_done_warm_up = False
+        start_time_lamp_warm_up = time.time()
+        while lamp_state.basicState != ATWhiteLight.LampBasicState.ON:
+            lamp_state = await self.white_light_source.evt_lampState.next(
+                flush=False,
+                timeout=self.timeout_lamp_warm_up
+            )
+            if lamp_state.basicState == ATWhiteLight.LampBasicState.WARMUP:
+                warmup_time_left = (
+                    lamp_state.warmupEndTime - lamp_state.private_rcvStamp
```
tribeiro, line 276: I think you probably want to use `private_sndStamp` instead of `private_rcvStamp`. Difference should be small but still.

python/lsst/ts/standardscripts/auxtel/calibrations/power_on_atcalsys.py
```diff
@@ -0,0 +1,330 @@
...
+                )
+                self.log.info(
+                    "Time Left for lamp warmup: {} min.".format(warmup_time_left / 60.0)
+                )
+                await asyncio.sleep(self.track_lamp_warmup)
```
tribeiro, line 281: drop the `self` from `self.track_lamp_warmup`, this is a local variable not a class attribute.

python/lsst/ts/standardscripts/auxtel/calibrations/power_on_atcalsys.py
```diff
@@ -0,0 +1,330 @@
...
+            elif lamp_state.basicState == ATWhiteLight.LampBasicState.ON:
+                lamp_done_warm_up = True
+                warm_up_elapsed_time = time.time() - start_time_lamp_warm_up
+                self.log.info(
+                    f"White Light Lamp is on after a warm up of {warm_up_elapsed_time} s"
+                )
+        else:
```
tribeiro, line 288: Is this correct? if the status is not `WARMUP` nor `ON` it will fail?

Also, if it is a fail, shouldn't you raise an exception or something?

python/lsst/ts/standardscripts/auxtel/calibrations/power_on_atcalsys.py
```diff
@@ -0,0 +1,330 @@
...
+            )
+            if lamp_state.basicState == ATWhiteLight.LampBasicState.WARMUP:
+                warmup_time_left = (
+                    lamp_state.warmupEndTime - lamp_state.private_rcvStamp
+                )
+                self.log.info(
+                    "Time Left for lamp warmup: {} min.".format(warmup_time_left / 60.0)
+                )
+                await asyncio.sleep(self.track_lamp_warmup)
+            elif lamp_state.basicState == ATWhiteLight.LampBasicState.ON:
```
tribeiro, line 282: I think you could drop this `elif` and just do this after the `while` loop has ended (it saves you from doing a comparison).

tribeiro (commented):
Here some initial comments.

python/lsst/ts/standardscripts/auxtel/calibrations/__init__.py
```diff
@@ -0,0 +1,21 @@
...
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+
```
tribeiro, line 20: Missing the final phrase:

```
# along with this program.  If not, see <https://www.gnu.org/licenses/>.
```

python/lsst/ts/standardscripts/auxtel/calibrations/power_on_atcalsys.py
```diff
@@ -0,0 +1,312 @@
...
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+
```
tribeiro, line 20:
```
# along with this program.  If not, see <https://www.gnu.org/licenses/>.
```

python/lsst/ts/standardscripts/auxtel/calibrations/power_on_atcalsys.py
```diff
@@ -0,0 +1,312 @@
...
+import yaml
+
+from lsst.ts.idl.enums import ATWhiteLight
+from lsst.ts.observatory.control.remote_group import Usages
+
+
+from lsst.ts import salobj
+
+track_lamp_warmup = 60
```
tribeiro, line 34: If this is supposed to be a constant, please, make is all caps, 

```
TRACK_LAMP_WARMUP = 60
```

But maybe best to have this as a class attribute instead, e.g.,. move it to the `__init__` method and make it `self.track_lamp_warmup`.

python/lsst/ts/standardscripts/auxtel/calibrations/power_on_atcalsys.py
```diff
@@ -0,0 +1,312 @@
...
+    """Powers on the ATCalSys dome flat illuminator
+    (ATWhiteLight and ATMonochromator)
+    required to perform image calibrations over white light.
+
+    Parameters
+    ----------
+    index : `int`
+        Index of Script SAL component.
+
+    Notes
```
tribeiro, line 47: If you don't have any Notes, you can remove this session.

python/lsst/ts/standardscripts/auxtel/calibrations/power_on_atcalsys.py
```diff
@@ -0,0 +1,312 @@
...
+
+        # White lamp config
+        self.timeout_lamp_warm_up = 60 * 15
+        self.cmd_timeout = 30
+
+        # Chiller config
+        self.timeout_chiller_cool_down = 60 * 10
+        self.chiller_temp_tolerance_relative = 0.1
+
+        if self.white_light_source is None:
```
tribeiro, line 69: I think you need to move all this code from here until line 86 to the `configure` method. Or maybe make a `setup_remotes` coroutine and await it in `configure`.

python/lsst/ts/standardscripts/auxtel/calibrations/power_on_atcalsys.py
```diff
@@ -0,0 +1,312 @@
...
+            self.white_light_source.start_task,
+            self.monochromator.start_task,
+        )
+
+
+    @classmethod
+    def get_schema(cls):
+        schema_yaml = """
+            $schema: http://json-schema.org/draft-07/schema#
+            $id: https://github.com/lsst-ts/ts_standardscripts/auxtel/SlewTelescopeIcrs.yaml
```
tribeiro, line 93: Fix id, should be something like:

```
https://github.com/lsst-ts/ts_standardscripts/python/lsst/standardscripts/auxtel/calibrations/power_on_atcalsys.py
```

you may have to break this down like we do [here](https://github.com/lsst-ts/ts_standardscripts/blob/f7285f06d5a178e24f84c28b3de7f62d376de47b/python/lsst/ts/standardscripts/maintel/m1m3/check_hardpoint.py#L70-L77).

python/lsst/ts/standardscripts/auxtel/calibrations/power_on_atcalsys.py
```diff
@@ -0,0 +1,312 @@
...
+        self.log.info("Configure started")
+
+        self.chiller_temperature = config.chiller_temperature
+        self.whitelight_power = config.whitelight_power
+        self.wavelength = config.wavelength
+        self.grating_type = config.grating_type
+        self.entrance_slit_width = config.entrance_slit_width
+        self.exit_slit_width = config.exit_slit_width
+
+        # if self.white_light_source is None:
```
tribeiro, line 162: Please, remove the commented code.. Also, move the code from __init__ to here...

python/lsst/ts/standardscripts/auxtel/calibrations/power_on_atcalsys.py
```diff
@@ -0,0 +1,312 @@
...
+            )
+
+        if self.monochromator is None:
+            self.monochromator = salobj.Remote(
+                domain=self.domain, name="ATMonochromator", 
+                intended_usage= None if add_remotes else Usages.DryTest, 
+                log=self.log
+            )
+
+        asyncio.gather(
```
tribeiro, line 83: You need to `await` this otherwise nothing happens.. Once you move this to the `configure` method, you will be able to do:

```
await asyncio.gather(...)
```

python/lsst/ts/standardscripts/auxtel/calibrations/power_on_atcalsys.py
```diff
@@ -0,0 +1,312 @@
...
+        #     self.white_light_source = salobj.Remote(
+        #         domain=self.domain, name="ATWhiteLight"
+        #     )
+
+        # if self.monochromator is None:
+        #     self.monochromator = salobj.Remote(
+        #         domain=self.domain, name="ATMonochromator"
+        #     )
+
+        self.config = config
```
tribeiro, line 172: Above you are creating individual attributes to get the values from `config` and then here you just store the entire thing. I think pick one method. Either set them up individually or get the full thing, but not both.

python/lsst/ts/standardscripts/auxtel/calibrations/power_on_atcalsys.py
```diff
@@ -0,0 +1,312 @@
...
+
+        Parameters
+        ----------
+        metadata : SAPY_Script.Script_logevent_metadataC
+        """
+        metadata.duration = self.timeout_chiller_cool_down + self.timeout_lamp_warm_up
+
+    async def run(self):
+        """Run script."""
+        #await self.assert_components_enabled()
```
tribeiro, line 187: Please, remove commented code or uncomment.

python/lsst/ts/standardscripts/auxtel/calibrations/power_on_atcalsys.py
```diff
@@ -0,0 +1,312 @@
...
+        await self.configure_atmonochromator()
+
+    async def start_chiller(self):
+        await self.white_light_source.cmd_setChillerTemperature.set_start(
+            temperature=self.chiller_temperature, timeout=self.cmd_timeout
+        )
+        await self.white_light_source.cmd_startChiller.set(timeout=self.timeout_chiller_cool_down)
+
+    async def wait_for_chiller_temp_within_tolerance(self):
+        start_time_chill_time = time.time()
```
tribeiro, line 214: time is repeated here.. maybe rename it `start_chill_time`?

python/lsst/ts/standardscripts/auxtel/calibrations/power_on_atcalsys.py
```diff
@@ -0,0 +1,312 @@
...
+        while time.time() - start_time_chill_time < self.timeout_chiller_cool_down:
+            chiller_temps = await self.white_light_source.tel_chillerTemperatures.aget()
+            tel_chiller_temp = chiller_temps.supplyTemperature
+            if (
```
tribeiro, line 218: Consider simplifying this condition to something like this:

```
            if (
                abs(chiller_temps.setTemperature - tel_chiller_temp)
                / chiller_temps.setTemperature
                <= self.chiller_temp_tolerance_relative
            ):
```

tribeiro (commented):
Here some comments for you to consider... I haven't been able to read through it all yet, sorry..

python/lsst/ts/standardscripts/auxtel/calibrations/power_on_atcalsys.py
isotuela: I have changed the code a little more.

python/lsst/ts/standardscripts/auxtel/calibrations/power_on_atcalsys.py
isotuela: I tried moving it to `__init__` because it didn't configure properly if it was in the `configure` method.

python/lsst/ts/standardscripts/auxtel/calibrations/power_on_atcalsys.py
isotuela: Thanks! I was testing different options.

python/lsst/ts/standardscripts/auxtel/calibrations/power_on_atcalsys.py
isotuela: Thanks!!

python/lsst/ts/standardscripts/auxtel/calibrations/power_on_atcalsys.py
isotuela: Uncommented.

python/lsst/ts/standardscripts/auxtel/calibrations/power_on_atcalsys.py
```diff
@@ -0,0 +1,294 @@
...
+    async def configure(self, config):
+        """Configure the script.
+
+        Parameters
+        ----------
+        config : ``self.cmd_configure.DataType``
+
+        Raises
+        ------
+        salobj.ExpectedError :
```
tribeiro, line 124: Usually we only document the exceptions that the method itself raises. This one in particular is raised by another method that does the configuration validation before calling `configure`. You should probably remove this here.

doc/version_history.rst
```diff
@@ -32,6 +32,8 @@ v1.22.0
 
 * Add new ``MoveP2P`` maintel script.
 
+* Add new ``auxtel/calibrations/power_on_atcalsys.py`` script, unit test and executable to set up and turn on the ATCalSys white light.
```
tribeiro, line 35: Version 1.22.0 was already released, please make this a new release 1.25.0

python/lsst/ts/standardscripts/auxtel/calibrations/power_on_atcalsys.py
```diff
@@ -0,0 +1,289 @@
...
+                continue
+        else:
+            raise TimeoutError(
+                f"Gave up waiting after {self.timeout_chiller_cool_down}"
+                f"for the chiller to chill to {self.chiller_temperature}."
+                f"Stayed at {tel_chiller_temp}"
+            )
+
+    async def open_white_light_shutter(self):
+        await self.white_light_source.cmd_openShutter.set_start()
```
tribeiro, line 218: Please, add a timeout to the command, otherwise it can stop here forever. Also, since you are not passing in any parameter, use, `start` instead, e.g,:

```
        await self.white_light_source.cmd_openShutter.start(timeout=self.timeout_open_shutter)
```

and you need to define `self.timeout_open_shutter` in the `__init__` method.

python/lsst/ts/standardscripts/auxtel/calibrations/power_on_atcalsys.py
```diff
@@ -0,0 +1,289 @@
...
+        # Q? Should we include a check that the shutter is open?
```
tribeiro, line 219: I think this command only finishes when the shutter is open, and fails if the shutter doesn't open.. So  think it is ok to leave it as is, e.g., no need to check.

python/lsst/ts/standardscripts/auxtel/calibrations/power_on_atcalsys.py
tribeiro: Sorry, not sure I understand the logic here. Do you need to break in this condition? it seems like it is going to continue in the loop until is time out. 

However, see my suggestion for simplifying this loop.

python/lsst/ts/standardscripts/auxtel/calibrations/power_on_atcalsys.py
```diff
@@ -0,0 +1,289 @@
...
+
+    async def switch_lamp_on(self):
+        await self.white_light_source.cmd_turnLampOn.set_start(
+            power=self.whitelight_power, timeout=self.timeout_lamp_warm_up
+        )
+
+    async def wait_for_lamp_to_warm_up(self):
+        self.white_light_source.evt_lampState.flush()
```
tribeiro, line 227: I think it is best to `flush` just before you send the command. Flushing here can create a race condition where the event is published between the time the command is sent and the time to get here, in which case you loose the event. 

I would actually suggest changing the loop a little bit. 

Consider something like this:

```
lamp_state = await self.white_light_source.evt_lampState.aget(
    timeout=self.timeout_lamp_warm_up
) # gets the most recent data

# Now enters the loop with the stop condition being that the lamp is ON

while lamp_state.basicState != ATWhiteLight.LampBasicState.ON:
    lamp_state = await self.white_light_source.evt_lampState.next(
        flush=False, timeout=self.timeout_lamp_warm_up
    )
    self.log.info(f"Lamp state: {ATWhiteLight.LampBasicState(lamp_state.basicState)!r}.")
```

python/lsst/ts/standardscripts/auxtel/calibrations/power_on_atcalsys.py
```diff
@@ -0,0 +1,289 @@
...
+
+        self.log.info(
+            f"ATMonochromator grating is {params[0]}, "
+            f"wavelength is {params[1]} nm "
+            f"with entry slit width {params[2]} "
+            f"and exit slit width {params[3]}"
+        )
+
+    async def get_monochromator_parameters(self):
+        tmp1 = await self.monochromator.evt_selectedGrating.aget()
```
tribeiro, line 276: Please, add a timeout for all these `aget` statements, otherwise it can block here for ever.

Also you can probably do this in "in parallel" using `asyncio.gather`:

```
return await asyncio.gather(
    self.monochromator.evt_selectedGrating.aget(timeout=self.std_timeout),
    self.monochromator.evt_wavelength.aget(timeout=self.std_timeout),
    self.monochromator.evt_entrySlitWidth.aget(timeout=self.std_timeout),
    self.monochromator.evt_exitSlitWidth.aget(timeout=self.std_timeout),
)
```

python/lsst/ts/standardscripts/auxtel/calibrations/power_on_atcalsys.py
```diff
@@ -0,0 +1,289 @@
...
+        return (tmp1.gratingType, tmp2.wavelength, tmp3.width, tmp4.width)
+
+    async def assert_components_enabled(self):
+        """Check if ATWhiteLight and ATMonochromator are ENABLED"""
+        for comp in [self.white_light_source, self.monochromator]:
+            summary_state = await comp.evt_summaryState.aget()
+            if salobj.State(summary_state.summaryState) != salobj.State(
+                salobj.State.ENABLED
+            ):
+                raise Exception(f"{comp} is not ENABLED")
```
tribeiro, line 289: Please, raise `RuntimeError` instead of a bare exception.

tribeiro (commented):
Here another round of comments..

python/lsst/ts/standardscripts/auxtel/calibrations/power_on_atcalsys.py
isotuela: I have put 5 minutes timeout, does that sound right?

python/lsst/ts/standardscripts/auxtel/calibrations/power_on_atcalsys.py
isotuela: Good suggestion.

PR #117, opened 2023-05-19
Tickets/sitcom 745

dsanmartim:
Hi @tribeiro ,

Please, take a look at this first draft of the actuators' bump test SAL script.

Tks,
David Sanmartim.

### DIFF
--- a/python/lsst/ts/standardscripts/data/scripts/maintel/m1m3/check_actuators.py
+++ b/python/lsst/ts/standardscripts/data/scripts/maintel/m1m3/check_actuators.py
@@ -1,3 +1,4 @@
+#!/usr/bin/env python
 # This file is part of ts_standardscripts
 #
 # Developed for the LSST Telescope and Site Systems.
@@ -17,9 +18,9 @@
 # GNU General Public License for more details.
 #
 # You should have received a copy of the GNU General Public License
-# along with this program. If not, see <https://www.gnu.org/licenses/>.
 
-from .check_hardpoint import *
-from .raise_m1m3 import *
-from .lower_m1m3 import *
-from .check_actuators import *
+import asyncio
+
+from lsst.ts.standardscripts.maintel.m1m3 import CheckActuators
+
+asyncio.run(CheckActuators.amain())
--- a/python/lsst/ts/standardscripts/maintel/m1m3/check_actuators.py
+++ b/python/lsst/ts/standardscripts/maintel/m1m3/check_actuators.py
@@ -1,0 +1,286 @@
+# This file is part of ts_standardscripts
+#
+# Developed for the LSST Telescope and Site Systems.
+# This product includes software developed by the LSST Project
+# (https://www.lsst.org).
+# See the COPYRIGHT file at the top-level directory of this distribution
+# for details of code ownership.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program. If not, see <https://www.gnu.org/licenses/>.
+
+__all__ = ["CheckActuators"]
+
+
+import time
+import yaml
+import asyncio
+
+from lsst.ts.salobj import BaseScript
+from lsst.ts.idl.enums.Script import ScriptState
+from lsst.ts.idl.enums.MTM1M3 import BumpTest, DetailedState
+from lsst.ts.observatory.control.maintel.mtcs import MTCS, MTCSUsages
+
+from lsst.ts.cRIOpy.M1M3FATable import (
+    FATABLE,
+    FATABLE_ID,
+)
+
+
+class CheckActuators(BaseScript):
+    """Perform a M1M3 bump test on either a selection of individual
+    actuators or on all actuators. Both cylinders will be tested for
+    dual actuators.
+
+    Parameters
+    ----------
+    index : `int`
+        Index of Script SAL component.
+
+    Notes
+    -----
+    **Checkpoints**
+
+    - "Running bump test on FA ID: {id}.": Check individual actuator.
+    - "M1M3 bump test completed.": Check complete.
+
+    """
+
+    def __init__(self, index, add_remotes: bool = True):
+        super().__init__(index=index, descr="Bump Test on M1M3 Actuators")
+
+        self.mtcs = MTCS(
+            domain=self.domain,
+            intended_usage=None if add_remotes else MTCSUsages.DryTest,
+            log=self.log,
+        )
+
+        # Average duration (seconds) of a bump test on a single actuator
+        self.time_one_bump = 25
+
+        # Getting list of actuator ids from mtcs
+        self.m1m3_actuator_ids = self.mtcs.get_m1m3_actuator_ids()
+        self.m1m3_secondary_actuator_ids = self.mtcs.get_m1m3_actuator_secondary_ids()
+
+        # Actuators that will be effectively tested
+        self.actuators_to_test = self.m1m3_actuator_ids.copy()
+
+    async def assert_feasibility(self):
+        """Verify that the system is in a feasible state before
+        running bump test.
+        """
+
+        for comp in self.mtcs.components_attr:
+            if comp != "mtm1m3":
+                self.log.debug(f"Ignoring component {comp}.")
+                setattr(self.mtcs.check, comp, False)
+
+        # Check all enabled, liveliness and the m1m3 detailed state
+        await asyncio.gather(
+            self.mtcs.assert_all_enabled(),
+            self.mtcs.assert_liveliness(),
+            self.mtcs.assert_m1m3_detailed_state(
+                {
+                    DetailedState.PARKED,
+                    DetailedState.PARKEDENGINEERING,
+                }
+            ),
+        )
+
+    @classmethod
+    def get_schema(cls):
+        m1m3_actuator_ids_str = ",".join([str(fa[FATABLE_ID]) for fa in FATABLE])
+
+        url = "https://github.com/lsst-ts/"
+        path = (
+            "ts_externalscripts/blob/main/python/lsst/ts/standardscripts/"
+            "maintel/m1m3/check_actuators.py"
+        )
+        schema_yaml = f"""
+        $schema: http://json-schema.org/draft-07/schema#
+        $id: {url}{path}
+        title: CheckAcutators v1
+        description: Configuration for Maintel bump test SAL Script.
+        type: object
+        properties:
+            actuators:
+                description: Actuators to run the bump test.
+                oneOf:
+                  - type: array
+                    items:
+                      type: number
+                      enum: [{m1m3_actuator_ids_str}]
+                    minItems: 1
+                    uniqueItems: true
+                    additionalItems: false
+                  - type: string
+                    enum: ["all"]
+                default: "all"
+        additionalProperties: false
+        """
+        return yaml.safe_load(schema_yaml)
+
+    async def configure(self, config):
+        """Configure the script.
+
+        Parameters
+        ----------
+        config : `types.SimpleNamespace`
+            Configuration
+        """
+
+        # Getting actuators to be tested.
+        self.actuators_to_test = (
+            self.m1m3_actuator_ids if config.actuators == "all" else config.actuators
+        )
+
+    def set_metadata(self, metadata):
+        """Set metadata."""
+
+        # Getting total number of secondary actuators to be tested
+        total_tested_secondary = sum(
+            [
+                1
+                for actuator in self.actuators_to_test
+                if self.has_secondary_actuator(actuator)
+            ]
+        )
+
+        # Setting metadata
+        metadata.duration = self.time_one_bump * (
+            len(self.actuators_to_test) + total_tested_secondary
+        )
+
+    def has_secondary_actuator(self, actuator_id: int) -> bool:
+        """Determines whether a given actuator has a
+        secondary axis or not.
+        """
+
+        return actuator_id in self.m1m3_secondary_actuator_ids
+
+    async def run(self):
+        await self.assert_feasibility()
+        start_time = time.monotonic()
+
+        # Get M1M3 detailed state
+        detailed_state = DetailedState(
+            (
+                await self.mtcs.rem.mtm1m3.evt_detailedState.aget(
+                    timeout=self.mtcs.fast_timeout,
+                )
+            ).detailedState
+        )
+        self.log.info(f"Current M1M3 detailed state: {detailed_state!r}.")
+
+        # Put M1M3 in engineering mode
+        await self.mtcs.enter_m1m3_engineering_mode()
+
+        failed_actuators_id = []
+        failed_primary, failed_secondary = [], []
+        for i, actuator_id in enumerate(self.actuators_to_test):
+            await self.mtcs.assert_all_enabled()
+
+            # Determine if actuator has a secondary actuator
+            secondary_exist = self.has_secondary_actuator(actuator_id)
+
+            # Checkpoint
+            primary_index = self.m1m3_actuator_ids.index(actuator_id)
+            if secondary_exist:
+                secondary_index = self.m1m3_secondary_actuator_ids.index(actuator_id)
+                await self.checkpoint(
+                    f"Running bump test on DAA FA ID: {actuator_id} "
+                    f"(Primary/Secondary: {primary_index}/{secondary_index})"
+                )
+            else:
+                await self.checkpoint(
+                    f"Running bump test on SAA FA ID: {actuator_id} "
+                    f"(Primary {primary_index})"
+                )
+
+            try:
+                await self.mtcs.run_m1m3_actuator_bump_test(
+                    actuator_id=actuator_id,
+                    primary=True,
+                    secondary=secondary_exist,
+                )
+            # raise a RuntimeError if the bump test fails
+            except RuntimeError:
+                failed_actuators_id.append(actuator_id)
+                self.log.exception(f"Failed to run bump test on FA ID {actuator_id}.")
+
+            # Getting test status
+            (
+                fa_id,
+                primary_status,
+                secondary_status,
+            ) = await self.mtcs.get_m1m3_bump_test_status()
+
+            secondary_status_text = (
+                f"Secondary FA (Index {secondary_index}): "
+                f"{secondary_status.name.upper()}."
+                if secondary_exist
+                else ""
+            )
+            self.log.info(
+                f"Bump test done for {i + 1} of {len(self.actuators_to_test)}. "
+                f"FA ID {fa_id}. Primary FA (Index {primary_index}): "
+                f"{primary_status.name.upper()}. "
+                f"{secondary_status_text}"
+            )
+
+            # Getting primary indexes for failures
+            if primary_status == BumpTest.FAILED:
+                failed_primary.append((actuator_id, primary_index))
+
+            if secondary_exist:
+                if secondary_status == BumpTest.FAILED:
+                    failed_secondary.append((actuator_id, secondary_index))
+
+        end_time = time.monotonic()
+        elapsed_time = (end_time - start_time) / 60.0
+
+        # Final checkpoint
+        await self.checkpoint(
+            f"M1M3 bump test completed. It took {elapsed_time:.2f} minutes"
+        )
+
+        # Final message with bump test results/status
+        if not failed_actuators_id:
+            self.log.info("All actuators PASSED the bump test.")
+        else:
+            failed_primary_str = ", ".join(
+                f"{failed[0]}:{failed[1]}" for failed in failed_primary
+            )
+            failed_secondary_str = ", ".join(
+                f"{failed[0]}:{failed[1]}" for failed in failed_secondary
+            )
+            error_message = (
+                f"Actuators {failed_actuators_id} FAILED the bump test. \n "
+                f"SAA (ID, Pri Index): {failed_primary_str} \n "
+                f"DAA (ID, Sec Index): {failed_secondary_str}"
+            )
+            self.log.error(error_message)
+            raise RuntimeError(error_message)
+
+    async def cleanup(self):
+        if self.state.state != ScriptState.ENDING:
+            try:
+                self.log.warning("M1M3 bump test stopped. Killing actuator forces.")
+
+                await self.mtcs.stop_m1m3_bump_test()
+
+            except Exception:
+                self.log.exception("Unexpected exception in stop_m1m3_bump_test.")
+
+        # Exiting engineering mode
+        await self.mtcs.exit_m1m3_engineering_mode()
--- a/tests/test_maintel_m1m3_check_actuators.py
+++ b/tests/test_maintel_m1m3_check_actuators.py
@@ -1,0 +1,174 @@
+# This file is part of ts_standardscripts
+#
+# Developed for the LSST Telescope and Site Systems.
+# This product includes software developed by the LSST Project
+# (https://www.lsst.org).
+# See the COPYRIGHT file at the top-level directory of this distribution
+# for details of code ownership.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program. If not, see <https://www.gnu.org/licenses/>.
+
+import asyncio
+import types
+import unittest
+
+from lsst.ts import salobj
+from lsst.ts.idl.enums.MTM1M3 import BumpTest, DetailedState
+
+from lsst.ts.standardscripts import get_scripts_dir
+from lsst.ts.standardscripts import BaseScriptTestCase
+from lsst.ts.standardscripts.maintel.m1m3 import CheckActuators
+
+
+class TestCheckActuators(BaseScriptTestCase, unittest.IsolatedAsyncioTestCase):
+    async def basic_make_script(self, index):
+        self.script = CheckActuators(index=index, add_remotes=False)
+        self.script.mtcs.run_m1m3_actuator_bump_test = unittest.mock.AsyncMock(
+            side_effect=self.mock_test_bump
+        )
+        self.script.mtcs.stop_m1m3_bump_test = unittest.mock.AsyncMock()
+        self.script.mtcs.enter_m1m3_engineering_mode = unittest.mock.AsyncMock()
+        self.script.mtcs.exit_m1m3_engineering_mode = unittest.mock.AsyncMock()
+        self.script.mtcs.assert_liveliness = unittest.mock.AsyncMock()
+        self.script.mtcs.assert_all_enabled = unittest.mock.AsyncMock()
+        self.script.mtcs.assert_m1m3_detailed_state = unittest.mock.AsyncMock()
+
+        self.script.mtcs.get_m1m3_bump_test_status = unittest.mock.AsyncMock(
+            side_effect=self.mock_get_m1m3_bump_test_status
+        )
+
+        self.script.mtcs.rem.mtm1m3 = unittest.mock.AsyncMock()
+        self.script.mtcs.rem.mtm1m3.configure_mock(
+            **{
+                "evt_detailedState.aget": self.get_m1m3_detailed_state,
+            }
+        )
+
+        self.bump_test_status = types.SimpleNamespace(
+            testState=[BumpTest.NOTTESTED] * len(self.script.m1m3_actuator_ids)
+        )
+
+        return (self.script,)
+
+    async def get_m1m3_detailed_state(self, *args, **kwags):
+        return types.SimpleNamespace(detailedState=DetailedState.PARKED)
+
+    # Side effects
+    async def mock_test_bump(self, actuator_id, primary, secondary):
+        await asyncio.sleep(0.5)
+        actuator_index = self.script.mtcs.get_m1m3_actuator_index(actuator_id)
+        self.bump_test_status.testState[actuator_index] = BumpTest.PASSED
+
+    # Create a side effect function for mock_bump_test_status method from mcts
+    # object. This function will be called when mock_bump_test_status method is
+    # called
+    async def mock_get_m1m3_bump_test_status(self):
+        self.m1m3_bump_test_status = 101, BumpTest.PASSED, BumpTest.PASSED
+        return self.m1m3_bump_test_status
+
+    async def test_configure_all(self):
+        """Testing a valid configuration: all actuators"""
+
+        # Configure with "all" actuators
+        async with self.make_script():
+            actuators = "all"
+
+            await self.configure_script(actuators=actuators)
+
+            assert self.script.actuators_to_test == self.script.m1m3_actuator_ids
+
+    async def test_configure_valid_ids(self):
+        """Testing a valid configuation: valid actuators ids"""
+
+        # Try configure with a list of valid actuators ids
+        async with self.make_script():
+            actuators = [101, 210, 301, 410]
+
+            await self.configure_script(
+                actuators=actuators,
+            )
+
+            assert self.script.actuators_to_test == actuators
+
+    async def test_configure_bad(self):
+        """Testing an invalid configuration: bad actuators ids"""
+
+        async with self.make_script():
+            # Invalid actuators: 501 and 505
+            actuators = [501, 505]
+
+            # If actuators_bad_ids is not empty, it should raise a ValueError
+            actuators_bad_ids = [
+                actuator
+                for actuator in actuators
+                if actuator not in self.script.m1m3_actuator_ids
+            ]
+            if actuators_bad_ids:
+                with self.assertRaises(salobj.ExpectedError):
+                    await self.configure_script(
+                        actuators=actuators_bad_ids,
+                    )
+
+    async def test_run(self):
+        # Run the script
+        async with self.make_script():
+            actuators = "all"
+            await self.configure_script(actuators=actuators)
+
+            # Run the script
+            await self.run_script()
+
+            # Assert all passed on mocked bump test. Had to get indexes.
+            actuators_to_test_index = [
+                self.script.mtcs.get_m1m3_actuator_index(actuator_id)
+                for actuator_id in self.script.actuators_to_test
+            ]
+
+            assert all(
+                [
+                    self.bump_test_status.testState[actuator_index] == BumpTest.PASSED
+                    for actuator_index in actuators_to_test_index
+                ]
+            )
+            # Expected awaint for assert_all_enabled method
+            expected_awaits = len(self.script.actuators_to_test) + 1
+
+            # Assert we await once for all mock methods defined above
+            self.script.mtcs.enter_m1m3_engineering_mode.assert_awaited_once()
+            self.script.mtcs.exit_m1m3_engineering_mode.assert_awaited_once()
+            self.script.mtcs.assert_liveliness.assert_awaited_once()
+            self.script.mtcs.assert_m1m3_detailed_state.assert_awaited_once()
+            assert self.script.mtcs.assert_all_enabled.await_count == expected_awaits
+
+            expected_calls = [
+                unittest.mock.call(
+                    actuator_id=actuator_id,
+                    primary=True,
+                    secondary=self.script.has_secondary_actuator(actuator_id),
+                )
+                for actuator_id in self.script.m1m3_actuator_ids
+            ]
+
+            self.script.mtcs.run_m1m3_actuator_bump_test.assert_has_calls(
+                expected_calls
+            )
+
+    async def test_executable(self):
+        scripts_dir = get_scripts_dir()
+        script_path = scripts_dir / "maintel" / "m1m3" / "check_actuators.py"
+        await self.check_executable(script_path)
+
+
+if __name__ == "__main__":
+    unittest.main()
--- a/doc/version_history.rst
+++ b/doc/version_history.rst
@@ -9,6 +9,7 @@ Version History
 v1.21.0
 -------
 
+* Add new ``maintel/m1m3/check_actuators.py`` script to run the actuators bump test.
 * Add new ``maintel/m1m3/lower_m1m3.py`` sal script and associated files. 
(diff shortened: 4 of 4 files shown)


python/lsst/ts/standardscripts/maintel/m1m3/check_actuators.py
```diff
@@ -0,0 +1,226 @@
...
+
+        # Check if actuators are valid
+        if not all([actuator in self.m1m3_actuator_table_id for actuator in self.actuators]):
+            raise ValueError("Invalid actuator ID.")
+
+    def set_metadata(self, metadata):
+        
+        metadata.duration = self.timeout_one_bump * (self.total_actuators + self.total_sec_actuators) 
+
+    def is_actuator_dual(self, actuator_id: int) -> bool:
```
tribeiro, line 149: Can you rename this method? Consider something like `is_actuator_secondary` or `has_secondary_actuator`. That is the nomenclature in the m1m3 interface.

python/lsst/ts/standardscripts/maintel/m1m3/check_actuators.py
```diff
@@ -0,0 +1,226 @@
...
+
+        for comp in self.mtcs.components_attr:
+            if comp != "mtm1m3":
+                self.log.debug(f"Ignoring component {comp}.")
+                setattr(self.mtcs.check, comp, False)
+
+        await asyncio.gather(
+            self.mtcs.assert_all_enabled(),
+            self.mtcs.assert_liveliness(),
+        )
```
tribeiro, line 91: Do you know if the m1m3 must be parked for this operations? if so, consider including

```
self.mtcs.assert_m1m3_detailed_state(
                {DetailedState.PARKED, DetailedState.PARKEDENGINEERING}
            )
```

here.

python/lsst/ts/standardscripts/maintel/m1m3/check_actuators.py
```diff
@@ -0,0 +1,226 @@
...
+        is_dual = FATABLE[actuator_index][M1M3FATable.FATABLE_TYPE] == 'DAA'
+
+        return is_dual
+
+    async def run(self):
+
+        await self.assert_feasibility()
+        start_time = time.time()
+
+        # Get M1M3 detailed state
```
tribeiro, line 163: I think you can skip logging the detailed state.

python/lsst/ts/standardscripts/maintel/m1m3/check_actuators.py
```diff
@@ -0,0 +1,226 @@
...
+        state = await self.mtcs.rem.mtm1m3.evt_detailedState.aget()
+        m1m3_detailed_state_name = self.MTM1M3.DetailedState(state.detailedState).name
+
+        self.log.info(f"M1M3 state before starting bump test: {m1m3_detailed_state_name}.")
+
```
tribeiro, line 168: I think you need to put m1m3 in engineering mode. You can call

```
await self.mtcs.enter_m1m3_engineering_mode()
```

it is a loop if the M1M3 is already in engineering mode so you don't need to do anything else.

python/lsst/ts/standardscripts/maintel/m1m3/check_actuators.py
```diff
@@ -0,0 +1,226 @@
...
+        self.log.info(f"M1M3 state before starting bump test: {m1m3_detailed_state_name}.")
+        
+        for i, actuator_id in enumarate(self.actuators): 
+            
+            await self.checkpoint(f"Running bump test on FA {actuator_id}.")
+          
+            # Determine if actuator has a secondary actuator
+            secondary_exist = self.is_actuator_dual(actuator_id)
+
+            if secondary_exist:
```
tribeiro, line 176: So, instead of this, consider simply doing:

```
await self.mtcs.run_m1m3_actuator_bump_test(
    actuator_id=actuator_id, 
    primary=True, 
    secondary=secondary_exist,
)
```

tribeiro (commented):
This looks pretty good. 

Here is a first round of comments for you to consider...

tribeiro (commented):
This looks pretty good. 

Here is a first round of comments for you to consider...

python/lsst/ts/standardscripts/maintel/m1m3/check_actuators.py
```diff
@@ -0,0 +1,226 @@
...
+        """
+
+        # Actuators defined as "all" (ID) or a list of actuators (INDEX)
+        self.actuators = (
+            self.m1m3_actuator_ids if config.actuators == "all" else config.actuators
+        )
+
+        # Check if actuators are valid
+        if not all([actuator in self.m1m3_actuator_table_id for actuator in self.actuators]):
+            raise ValueError("Invalid actuator ID.")
```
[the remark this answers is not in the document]
dsanmartim: I think it is better to keep it, in case the schema does not work as expected. Added your suggestion.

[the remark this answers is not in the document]
dsanmartim: BTW... I'm checking the "Resolve conversation" button after addressing comments, but not sure if the project or team has guidelines/practices established for handling and resolving PR comments. Is this ok?

python/lsst/ts/standardscripts/maintel/m1m3/check_actuators.py
dsanmartim: The original idea of this script according to its ticket ([SITCOM-745](https://jira.lsstcorp.org/browse/SITCOM-745)) was:

_The SAL Script will consider only the Bump Test itself. It should check if M1M3 is in ACTIVEENGINEERING or in PARKEDENGINEERING detailed state, but it should NOT change the detailed state. This way, we can have a single script to run both test cases and avoid duplication._

That is why I'm logging the M1M3 detailed state here...

python/lsst/ts/standardscripts/maintel/m1m3/check_actuators.py
dsanmartim: How is this related to the comment above? Will this command put the m1m3 in parked on active engineering mode?

python/lsst/ts/standardscripts/maintel/m1m3/check_actuators.py
dsanmartim: Ohhh... I just saw this after answering a similar question below. The original idea of this script was to test it either in `DetailedState.PARKEDENGINEERING` or `DetailedState.ACTIVEENGINEERING` but without changing the detailed state. However, the reason for that is to avoid script duplication. That could be avoided by modifying the JSON schema to include those options, right?

dsanmartim: In any case, I'm adding the following to the code:

```
self.mtcs.assert_m1m3_detailed_state(
            {
                DetailedState.PARKED,
                DetailedState.PARKEDENGINEERING,
                DetailedState.ACTIVE,
                DetailedState.ACTIVEENGINEERING,
            }
        )
```
Is this correct?

Also: flake8 is complaining about the detailed states, saying e.g. that the `DetailedState` name is undefined. Should I import something?

tribeiro: yes, you need to import it:

```
from lsst.ts.idl.enums.MTM1M3 import DetailedState
```

python/lsst/ts/standardscripts/maintel/m1m3/check_actuators.py
tribeiro: If you do the check of the detailed state in `asset_feasibility` it will raise an exception when the detailed state is not the desired one, and continue if it is ok. I think this is enough information.

If you insist in logging it, please add a timeout in `aget` to avoid blocking here for ever if there's ever loss of historical data.

tribeiro (commented):
Hear some additional comments/responses. 

Did you pushed your latest changes? I see some conversations marked as resolved and some 👍 , but the code looks the same as before.

python/lsst/ts/standardscripts/maintel/m1m3/check_actuators.py
tribeiro: I think you need to remove this `if` statement. otherwise the test won't run if `secondary_exist` is `False`

dsanmartim: Yeah, I was just reviewing that. Thanks for catching this too!

python/lsst/ts/standardscripts/maintel/m1m3/check_actuators.py
```diff
@@ -0,0 +1,266 @@
...
+                self.log.debug(f"Ignoring component {comp}.")
+                setattr(self.mtcs.check, comp, False)
+
+        await asyncio.gather(
+            self.mtcs.assert_all_enabled(),
+            self.mtcs.assert_liveliness(),
+        )
+
+        # Check the m1m3 detailed state
+        self.mtcs.assert_m1m3_detailed_state(
```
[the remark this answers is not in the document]
dsanmartim: Ok. I put it in the `gather` above!

python/lsst/ts/standardscripts/maintel/m1m3/check_actuators.py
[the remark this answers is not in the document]
dsanmartim: I got it. I'm about to start writing a unit test!

python/lsst/ts/standardscripts/maintel/m1m3/check_actuators.py
```diff
@@ -0,0 +1,245 @@
...
+    async def configure(self, config):
+        """Configure the script.
+
+        Parameters
+        ----------
+        config : `types.SimpleNamespace`
+            Configuration
+        """
+
+        # Getting actuators that to be tested.
```
tribeiro, line 143: typo, remove "that", 

> Getting actuators to be tested.

python/lsst/ts/standardscripts/maintel/m1m3/check_actuators.py
```diff
@@ -0,0 +1,245 @@
...
+        )
+
+    async def cleanup(self):
+        if self.state.state != ScriptState.ENDING:
+            try:
+                self.log.warning("M1M3 bump test stopped. Killing actuator forces.")
+
+                await self.mtcs.stop_m1m3_bump_test()
+
+            except asyncio.TimeoutError:
```
tribeiro, line 240: I think you can skip catching `TimeoutError` and just use the global `Exception` below.

python/lsst/ts/standardscripts/maintel/m1m3/check_actuators.py
```diff
@@ -0,0 +1,245 @@
...
+                self.log.exception(
+                    "Stopping bump test command timed out during cleanup procedure."
+                )
+            except Exception:
+                self.log.exception("Unexpected exception in stop_m1m3_bump_test.")
```
tribeiro, line 245: Remember to exit engineering state at the end with:

```
await self.mtcs.exit_m1m3_engineering_mode()
```

tribeiro (commented):
This is looking pretty good now! 

I think the missing parts are to add the unit tests for the script and also handle test failures. 

I left a few additional comments inline.

python/lsst/ts/standardscripts/maintel/m1m3/check_actuators.py
```diff
@@ -0,0 +1,252 @@
...
+                )
+
+            else:
+
+                self.log.info(
+                    f"Bump test on FA ID {fa_id}. Primary/Secondary: "
+                    f"{primary_status.upper()}/{secondary_status.upper()} ."
+                )
+
+            self.log.info(f"Bump test done for {i + 1} of {len(self.actuators_to_test)}")
```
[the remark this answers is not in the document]
dsanmartim: I moved it to the above message, as I think it gives an idea of the test progress.

python/lsst/ts/standardscripts/maintel/m1m3/check_actuators.py
dsanmartim: Is it ok giving a `log.info` message when entering/exiting engineering mode?

python/lsst/ts/standardscripts/maintel/m1m3/check_actuators.py
```diff
@@ -0,0 +1,261 @@
...
+                    primary=True,
+                    secondary=secondary_exist,
+                )
+            # raise a RuntimeError if the bump test fails
+            # and continue loop to next actuator_id
+            except RuntimeError as e:
+                failed_actuators.append(actuator_id)
+
+                self.log.error(
+                    f"Failed to run bump test on FA ID {actuator_id}." f"Error: {e}"
```
tribeiro, line 208: you can merge the 2 strings into 1 apparently:

```
f"Failed to run bump test on FA ID {actuator_id}. Error: {e}"
```

However, if you want to log the exception, use `self.log.exception` instead:

```
except RuntimeError:
    failed_actuators.append(actuator_id)
    self.log.exception(f"Failed to run bump test on FA ID {actuator_id}.")
```

python/lsst/ts/standardscripts/maintel/m1m3/check_actuators.py
tribeiro: The [mtcs](https://github.com/lsst-ts/ts_observatory_control/blob/d94417554a84c4384921042f82b38d23aac88967/python/lsst/ts/observatory/control/maintel/mtcs.py#L1033) already logs it...

python/lsst/ts/standardscripts/maintel/m1m3/check_actuators.py
```diff
@@ -0,0 +1,257 @@
...
+                    f"Bump test on FA ID {fa_id}. Primary/Secondary: "
+                    f"{primary_status.upper()}/{secondary_status.upper()} ."
+                )
+
+        end_time = time.time()
+        elapsed_time = (end_time - start_time) / 60.0
+
+        # Final message with bump test results/status
+        if not failed_actuators:
+            await self.log.info("All actuators PASSED the bump test.")
```
tribeiro, line 236: remove `await` before `self.log.info`

python/lsst/ts/standardscripts/maintel/m1m3/check_actuators.py
```diff
@@ -0,0 +1,257 @@
...
+        else:
+            await self.log.warning(
```
tribeiro, line 238: You probably want to fail the Script in this case right?

In this case, raise a `RuntimeError` exception here, e.g.:

```
error_message = f"Actuators {failed_actuators} FAILED the bump test."
self.log.error(error_message)
raise RuntimeError(error_message)
```

python/lsst/ts/standardscripts/maintel/m1m3/check_actuators.py
```diff
@@ -0,0 +1,257 @@
...
+                f"Actuators {failed_actuators} FAILED the bump test. "
+            )
+        # Final checkpoint
+        await self.checkpoint(
```
tribeiro, line 242: You may want to wait for this checkpoint before the if/else above because we probably want to raise an exception if the test fail, but still want to have this checkpoint.

tribeiro (commented):
Alright, this is getting pretty close. 

I think it is probably time to squash all these commits and update the version history.

python/lsst/ts/standardscripts/maintel/m1m3/check_actuators.py
```diff
@@ -0,0 +1,258 @@
...
+
+        # Check all enabled, liveliness and the m1m3 detailed state
+        await asyncio.gather(
+            self.mtcs.assert_all_enabled(),
+            self.mtcs.assert_liveliness(),
+            self.mtcs.assert_m1m3_detailed_state(
+                {
+                    DetailedState.PARKED,
+                    DetailedState.PARKEDENGINEERING,
+                    DetailedState.ACTIVE,
```
pkubanek, line 95: You cannot (yet) run the bump test on the active mirror. Note T018 is a different beast. If we want that to run on the active mirror (and given the CSC can run that probably better, plus we will be able to run a bump test from EUI, which I am unable to do now), I need a ticket for M1M3.

You can still run T018-like bump test, but that will require you to implement T018-like logic, which is a lot more difficult than simply firing bump_test_FA.

I suggest leaving ACTIVE (and ACTIVEENGINEERING) out now. The problem with bump testing while active is that it will not tell you anything about whatever a FA is performing as expected - you need to raise the mirror first, and if an FA is misbehaving, it might be too late to save the mirror as you are rising it. Running bump test with mirror active is like using a tool to verify that a function (raising the mirror) can proceed after it already finished.

We now run T018 during mirror verification, but that's mainly to see FAs works at elevation. 002, the bump test while lowered, will do a better job. I don't envision why we shall run bump test when active during operation to verify FA performance.

python/lsst/ts/standardscripts/maintel/m1m3/check_actuators.py
```diff
@@ -0,0 +1,258 @@
...
+        description: Configuration for Maintel bump test SAL Script.
+        type: object
+        properties:
+            actuators:
+                description: Actuators to run the bump test.
+                oneOf:
+                  - type: array
+                    items:
+                      type: number
+                      enum: [{m1m3_actuator_ids_str}]
```
pkubanek, line 123: How you distinguish between primary and secondary FAs? It might be a valid case. For the moment, add to description something like "Both cylinders will be tested for DAA."

python/lsst/ts/standardscripts/maintel/m1m3/check_actuators.py
```diff
@@ -0,0 +1,258 @@
...
+    def has_secondary_actuator(self, actuator_id: int) -> bool:
+        """Determines whether a given actuator has a
+        secondary axis or not.
+        """
+
+        return actuator_id in self.m1m3_secondary_actuator_ids
+
+    async def run(self):
+        await self.assert_feasibility()
+        start_time = time.time()
```
pkubanek, line 174: Please avoid time.time() for time measurement. Use time.monotonic() instead.

python/lsst/ts/standardscripts/maintel/m1m3/check_actuators.py
```diff
@@ -0,0 +1,258 @@
...
+        )
+        self.log.info(f"Current M1M3 detailed state: {detailed_state!r}.")
+
+        # Put M1M3 in engineering mode
+        await self.mtcs.enter_m1m3_engineering_mode()
+
+        failed_actuators = []
+        for i, actuator_id in enumerate(self.actuators_to_test):
+            await self.mtcs.assert_all_enabled()
+            await self.checkpoint(f"Running bump test on FA ID: {actuator_id}.")
```
pkubanek, line 192: For digging out data from EFD, it's handy to have indices in the log message. Consider something like:
```python
    primary_index = self.m1m3_actuator_ids.index(actuator_id)
    if self.has_secondary(actuator_id):
        secondary_index = self.m1m3_secondary_actuator_ids.index()
        await self.checkpoint(f"Running bump test on DAA FA ID: {actuator_id} (primary {primary_index} secondary {secondary_index})")
    else:
        await self.checkpoint(f"Running bump test on SAA FA ID: {actuator_id} (primary {primary_index})")
```

python/lsst/ts/standardscripts/maintel/m1m3/check_actuators.py
```diff
@@ -0,0 +1,258 @@
...
+            try:
+                await self.mtcs.run_m1m3_actuator_bump_test(
+                    actuator_id=actuator_id,
+                    primary=True,
+                    secondary=secondary_exist,
+                )
+            # raise a RuntimeError if the bump test fails
+            # and continue loop to next actuator_id
+            except RuntimeError:
+                failed_actuators.append(actuator_id)
```
pkubanek, line 206: Although API allows for bump test commands to run both bump tests, it's quite useful to know at the end which cylinder (primary or secondary) faulted. Please have a look at T002, which implements that by separating primary/secondary bump tests.

python/lsst/ts/standardscripts/maintel/m1m3/check_actuators.py
```diff
@@ -0,0 +1,258 @@
...
+                fa_id,
+                primary_status,
+                secondary_status,
+            ) = await self.mtcs.get_m1m3_bump_test_status()
+
+            if not secondary_exist:
+                self.log.info(
+                    f"Bump test done for {i + 1} of {len(self.actuators_to_test)}"
+                    f"Bump test on FA ID {fa_id}. Primary FA: "
+                    f"{primary_status.upper()}."
```
pkubanek, line 221: If the primary_status is FAILED, you shall add that FA to failed FAs. This might negate need for running separate bump tests mentioned above.

pkubanek (changes requested):
Please at least remove ACTIVE and ACTIVEENGINEERING states from the allowed states.

python/lsst/ts/standardscripts/maintel/m1m3/check_actuators.py
dsanmartim: I would prefer to include this information in the final message, along with the bump test results.

python/lsst/ts/standardscripts/maintel/m1m3/check_actuators.py
dsanmartim: I took the `self.mtcs.get_m1m3_bump_test_status()` and included the indexes in the final message.

python/lsst/ts/standardscripts/maintel/m1m3/check_actuators.py
```diff
@@ -0,0 +1,284 @@
...
+            # Getting test status
+            (
+                fa_id,
+                primary_status,
+                secondary_status,
+            ) = await self.mtcs.get_m1m3_bump_test_status()
+
+            if not secondary_exist:
+                # Getting primary indexes for failures
+                if primary_status == "BumpTest.FAILED":
```
pkubanek, line 235: This doesn't return a string. This will not work - please remove ":

```
In [3]: MTM1M3.BumpTest.FAILED == "BumpTest.FAILED"
Out[3]: False
```

python/lsst/ts/standardscripts/data/scripts/maintel/m1m3/check_actuators.py
```diff
@@ -0,0 +1,26 @@
...
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
```
gmegh, line 20: Missing comment line: "# along with this program. If not, see <https://www.gnu.org/licenses/>."

python/lsst/ts/standardscripts/maintel/m1m3/check_actuators.py
dsanmartim: Hi Petr. Thanks for catching this. I tested again with `if str(primary_status) == 'BumpTest. FAILED'` and it does work. I have changed it to str(primary_status).

python/lsst/ts/standardscripts/data/scripts/maintel/m1m3/check_actuators.py
dsanmartim: Now that you comment on this, I see that lots of scripts on this repo have that line missing. However, I see that line on the `check_actuators.py`. Let me know if you still don't see it.

gmegh: It's in the check_actuators.py but not in the data/script check_actuators.py i believe. For the other files, we are currently updating it on another PR

pkubanek: https://github.com/lsst-ts/ts_standardscripts/pull/118

python/lsst/ts/standardscripts/maintel/m1m3/check_actuators.py
```diff
@@ -0,0 +1,284 @@
...
+        # Final checkpoint
+        await self.checkpoint(
+            f"M1M3 bump test completed. It took {elapsed_time:.2f} minutes"
+        )
+
+        # Final message with bump test results/status
+        if not failed_actuators:
+            self.log.info("All actuators PASSED the bump test.")
+        else:
+            error_message = f"Actuators {failed_actuators} FAILED the bump test. \n "
```
tribeiro, line 268: You can build this error message without adding the strings with

```
            error_message = (
                f"Actuators {failed_actuators} FAILED the bump test. \n "
                f"Primary Indexes: {failed_primary_index} \n "
                f"Secondary Indexes: {failed_secondary_index}"
            )
```

tests/test_maintel_m1m3_check_actuators.py
```diff
@@ -0,0 +1,128 @@
...
+
+    async def test_configure_bad(self):
+        """Testing an invalid configuration"""
+
+        async with self.make_script():
+            # Invalid actuators: 501 and 505
+            actuators = [101, 210, 301, 410, 501, 505]
+
+            # Check if actuators are valid
+            # if not all([actuator in self.m1m3_actuator_ids for actuator
```
tribeiro, line 87: Please, remove commented code..

python/lsst/ts/standardscripts/maintel/m1m3/check_actuators.py
dsanmartim: Thanks, Petr. I'm using the enums now, as you suggested: `if MTM1M3.BumpTest. FAILED == primary_status`

tests/test_maintel_m1m3_check_actuators.py
```diff
@@ -0,0 +1,128 @@
...
+            )
+
+            assert self.script.actuators == actuators
+
+    async def test_configure_bad(self):
+        """Testing an invalid configuration"""
+
+        async with self.make_script():
+            # Invalid actuators: 501 and 505
+            actuators = [101, 210, 301, 410, 501, 505]
```
tribeiro, line 84: I think you can test the bad ids only

tests/test_maintel_m1m3_check_actuators.py
```diff
@@ -0,0 +1,128 @@
...
+
+from lsst.ts.standardscripts import get_scripts_dir
+from lsst.ts.standardscripts import BaseScriptTestCase
+from lsst.ts.standardscripts.maintel.m1m3 import CheckActuators
+
+
+class TestCheckActuators(BaseScriptTestCase, unittest.IsolatedAsyncioTestCase):
+    async def basic_make_script(self, index):
+        self.script = CheckActuators(index=index, add_remotes=False)
+        self.script.mtcs.run_m1m3_bump_test = unittest.mock.AsyncMock(
```
tribeiro, line 36: I think this should be

```
self.script.mtcs.run_m1m3_actuator_bump_test
```

no?

python/lsst/ts/standardscripts/maintel/m1m3/check_actuators.py
```diff
@@ -0,0 +1,285 @@
...
+            # Getting test status
+            (
+                fa_id,
+                primary_status,
+                secondary_status,
+            ) = await self.mtcs.get_m1m3_bump_test_status()
+
+            # BumpTest.PASSED = 6
+            # BumpTest.FAILED = 7
+            if not secondary_exist:
```
pkubanek, line 234: This will not add failed primary if there is secondary. There is always primary. So the test shall be:

```
               # Getting primary indexes for failures
                if primary_status == BumpTest.FAILED:
                    failed_primary_index.append(primary_index)
                if secondary_exists:
                    if secondary_status == BumpTest.FAILED:
                        failed_secondary_index.append(secondary_index)
```

but even better (to identify), if you append both index and id, like that:
```
failed_secondary.append(actuator_id, secondary_index)
```

python/lsst/ts/standardscripts/maintel/m1m3/check_actuators.py
```diff
@@ -0,0 +1,285 @@
...
+            f"M1M3 bump test completed. It took {elapsed_time:.2f} minutes"
+        )
+
+        # Final message with bump test results/status
+        if not failed_actuators:
+            self.log.info("All actuators PASSED the bump test.")
+        else:
+            error_message = f"Actuators {failed_actuators} FAILED the bump test. \n "
+            error_message += f"Primary Indexes: {failed_primary_index} \n "
+            error_message += f"Secondary Indexes: {failed_secondary_index}"
```
pkubanek, line 270: The message is a bit misleading, mixing IDs and indices. This might cause a lot of troubles looking for the actuator. See above for one option.

pkubanek (changes requested):
You always have primary, so need to see it it passed.

tests/test_maintel_m1m3_check_actuators.py
```diff
@@ -0,0 +1,128 @@
...
+
+            await self.run_script()
+
+            assert all(
+                [
+                    self.script.bump_test_status.testState[actuator] == BumpTest.PASSED
+                    for actuator in self.script.actuators
+                ]
+            )
+
```
tribeiro, line 120: I think you could do some additional assertions about the expected calls done in the method you mocked.

For example, you expect `self.script.mtcs.enter_m1m3_engineering_mode` to be awaited once, so you can add

```
self.script.mtcs.enter_m1m3_engineering_mode.assert_awaited_once()
```

and the same for all the other methods that need to be awaited once. 

Also, when you run with `actuators="all"` you expect `self.script.mtcs.run_m1m3_actuator_bump_test` to be awaited with all the hp `actuator_id` and the appropriate `primary` and `secondary` values. 

you can do something like this:

```
expected_calls = [
    unittest.mock.call(
        actuator_id=actuator_id,
        primary=True,
        secondary=self.script.has_secondary_actuator(actuator_id)
    ) for actuator_id in self.script.m1m3_actuator_ids
]

self.script.mtcs.run_m1m3_actuator_bump_test.assert_has_calls(expected_calls)
```

tribeiro (commented):
It's looking pretty good.. Please, address Petr's comments and the additional few I am leaving.. I think after that we will be good to merge.

tests/test_maintel_m1m3_check_actuators.py
dsanmartim: Good catch!

python/lsst/ts/standardscripts/maintel/m1m3/check_actuators.py
dsanmartim: Hi Petr. One question here. This will only check the m1m3 state (raised or lowered), without changing its state. So, considering that we will use this script to run bump tests in both states (either for mirror lowered and raised, which will be done using the m1m3_lower and m1m3_raise), we should keep this checking, don't you think?

pkubanek: Keep in only PARKED and PARKEDENGINEERING states. We will handle ACTIVE states later (if needed).

tests/test_maintel_m1m3_check_actuators.py
```diff
@@ -0,0 +1,142 @@
...
+                with self.assertRaises(ValueError):
+                    await self.configure_script(
+                        actuators=actuators_bad_ids,
+                    )
+
+    async def test_run(self):
+        # Run the script
+        async with self.make_script():
+            # Assert feasibility
+            self.script.mtcs.assert_feasibility()
```
tribeiro, line 102: Please, remove this call

tests/test_maintel_m1m3_check_actuators.py
```diff
@@ -0,0 +1,142 @@
...
+
+            # Await once for all mock methods defined on the basic_make_script
+            self.script.mtcs.enter_m1m3_engineering_mode.assert_awaited_once()
```
tribeiro, line 105: You have to make these assertions after you run the test, otherwise they will not have been called.

tests/test_maintel_m1m3_check_actuators.py
```diff
@@ -0,0 +1,142 @@
...
+            self.script.mtcs.exit_m1m3_engineering_mode.assert_awaited_once()
+            self.script.mtcs.assert_liveliness.assert_awaited_once()
+            self.script.mtcs.assert_all_enabled.assert_awaited_once()
+            self.script.mtcs.assert_m1m3_detailed_state.assert_awaited_once()
+            self.script.mtcs.assert_feasibility.assert_awaited_once()
```
tribeiro, line 110: This doesn't exist. You might be thinking of mocking `self.script.assert_feasibility`, but you shouldn't. Mock only external stuff (methods from mtcs).

tests/test_maintel_m1m3_check_actuators.py
```diff
@@ -0,0 +1,142 @@
...
+        self.script = CheckActuators(index=index, add_remotes=False)
+        self.script.mtcs.run_m1m3_actuator_bump_test = unittest.mock.AsyncMock(
+            side_effect=self.mock_test_bump
+        )
+        self.script.mtcs.enter_m1m3_engineering_mode = unittest.mock.AsyncMock()
+        self.script.mtcs.exit_m1m3_engineering_mode = unittest.mock.AsyncMock()
+        self.script.mtcs.assert_liveliness = unittest.mock.AsyncMock()
+        self.script.mtcs.assert_all_enabled = unittest.mock.AsyncMock()
+        self.script.mtcs.assert_m1m3_detailed_state = unittest.mock.AsyncMock()
+        self.script.mtcs.assert_feasibility = unittest.mock.AsyncMock()
```
tribeiro, line 44: This doesn't exist. You might be thinking of mocking `self.script.assert_feasibility`, but you shouldn't. Mock only external stuff (methods from mtcs).

tests/test_maintel_m1m3_check_actuators.py
tribeiro: `self.script.mtcs.assert_feasibility` doesn't exist and you should not mock `self.script.assert_feasibility`

tests/test_maintel_m1m3_check_actuators.py
```diff
@@ -0,0 +1,138 @@
...
+            side_effect=self.mock_test_bump
+        )
+        self.script.mtcs.enter_m1m3_engineering_mode = unittest.mock.AsyncMock()
+        self.script.mtcs.exit_m1m3_engineering_mode = unittest.mock.AsyncMock()
+        self.script.mtcs.assert_liveliness = unittest.mock.AsyncMock()
+        self.script.mtcs.assert_all_enabled = unittest.mock.AsyncMock()
+        self.script.mtcs.assert_m1m3_detailed_state = unittest.mock.AsyncMock()
+
+        self.bump_test_status = types.SimpleNamespace(
+            testState=[BumpTest.NOTTESTED] * len(self.m1m3_actuator_ids)
```
tribeiro, line 46: Should this be:

```
self.bump_test_status = types.SimpleNamespace(
             testState=[BumpTest.NOTTESTED] * len(self.script.m1m3_actuator_ids)
```

instead?

meaning, replace `self.m1m3_actuator_ids` with `self.script.m1m3_actuator_ids`

tests/test_maintel_m1m3_check_actuators.py
```diff
@@ -0,0 +1,138 @@
...
+        # Configure with "all" actuators
+        async with self.make_script():
+            actuators = "all"
+
+            await self.configure_script(actuators=actuators)
+
+            assert self.script.actuators == self.m1m3_actuator_ids
+
+        # Try configure with a list of valid actuators ids
+        async with self.make_script():
```
tribeiro, line 67: Consider spitting each of these `async with` in its own test.. It improves the testing reporting.

e.g.:

```
        async def test_configure_all(self):

         # Configure with "all" actuators
             actuators = "all"

             await self.configure_script(actuators=actuators)

             assert self.script.actuators == self.m1m3_actuator_ids

    async def test_configure_valid_ids(self):
            actuators = [101, 210, 301, 410]

             await self.configure_script(
                 actuators=actuators,
             )

             assert self.script.actuators == actuators
```

tribeiro (approved):
I only have a couple comments in the testing case. I think other than that everything else is good to go

tests/test_maintel_m1m3_check_actuators.py
dsanmartim: Good eyes!

python/lsst/ts/standardscripts/maintel/m1m3/check_actuators.py
```diff
@@ -0,0 +1,285 @@
...
+                fa_id,
+                primary_status,
+                secondary_status,
+            ) = await self.mtcs.get_m1m3_bump_test_status()
+
+            # Getting primary indexes for failures
+            if primary_status == BumpTest.FAILED:
+                failed_primary.append(actuator_id, primary_index)
+
+                self.log.info(
```
pkubanek, line 232: You probably want to move that before test for FAILED. Also for secondary. That way it will report only on failed FAs. Or better yet, remove those logs, and add status to # final checkpoint.

python/lsst/ts/standardscripts/maintel/m1m3/check_actuators.py
```diff
@@ -0,0 +1,285 @@
...
+        # Final checkpoint
+        await self.checkpoint(
+            f"M1M3 bump test completed. It took {elapsed_time:.2f} minutes"
+        )
+
+        # Final message with bump test results/status
+        if not failed_actuators_id:
+            self.log.info("All actuators PASSED the bump test.")
+        else:
+            failed_primary_str = " ".join(
```
pkubanek, line 260: maybe better:
```
failed_primary_str = " ".join([f"{failed[0]}:{failed[1]} for failed in failed_primary"])
```
the use of str(item).replace is a bit non-standard, if Python changed str repr of tuple, it will ceased to work

pkubanek (approved):
Please see suggested changes.

python/lsst/ts/standardscripts/maintel/m1m3/check_actuators.py
dsanmartim: I think this is not going to work. Actually, the `failed_primary` list has the following format: `[(101, 1), (102, 2), (103, 3)]`. When we do the `failed_primary.append(actuator_id, index)`, we will get a failure, as the `append` method can only accept one argument, so we have to append tuples to the list, like this `failed_primary.append((actuator_id, index))`

python/lsst/ts/standardscripts/maintel/m1m3/check_actuators.py
dsanmartim: Good catch. I removed it from the `if primary_status == BumpTest.FAILED` statement. But for now, I have kept the message here, to give information while we progress with the bump test.

python/lsst/ts/standardscripts/maintel/m1m3/check_actuators.py
```diff
@@ -0,0 +1,286 @@
...
+        )
+
+    def has_secondary_actuator(self, actuator_id: int) -> bool:
+        """Determines whether a given actuator has a
+        secondary axis or not.
+        """
+
+        return actuator_id in self.m1m3_secondary_actuator_ids
+
+    async def arun(self, checkpoint=False):
```
tribeiro, line 171: Please remove this workaround!

tribeiro (changes requested):
Please, remove the workaround you included to run it at TTS.

dsanmartim:
Thank you all for your comments. Finally merged!

PR #121, opened 2023-06-05
Tickets/sitcom 884

dsanmartim:
Hi Tiago, 

Here are a few changes to the `check_actuators.py` script. More details are in the corresponding Jira ticket.

Tks!

### DIFF
--- a/doc/version_history.rst
+++ b/doc/version_history.rst
@@ -9,6 +9,7 @@ Version History
 v1.22.0
 -------
 
+* Update the ``maintel/m1m3/check_actuators.py`` script with improved logging and detailed state assertions.
 * Add new ``maintel/home_both_axes.py`` script to home both MTMount axes. 
 
 
--- a/python/lsst/ts/standardscripts/maintel/m1m3/check_actuators.py
+++ b/python/lsst/ts/standardscripts/maintel/m1m3/check_actuators.py
@@ -39,8 +39,7 @@ from lsst.ts.cRIOpy.M1M3FATable import (
 
 class CheckActuators(BaseScript):
     """Perform a M1M3 bump test on either a selection of individual
-    actuators or on all actuators. Both cylinders will be tested for
-    dual actuators.
+    actuators or on all actuators.
 
     Parameters
     ----------
@@ -49,6 +48,10 @@ class CheckActuators(BaseScript):
 
     Notes
     -----
+
+    In case of dual actuators both cylinders will be tested consecutively.
+    The Script will fail if M1M3 mirror is raised.
+
     **Checkpoints**
 
     - "Running bump test on FA ID: {id}.": Check individual actuator.
@@ -77,7 +80,8 @@ class CheckActuators(BaseScript):
 
     async def assert_feasibility(self):
         """Verify that the system is in a feasible state before
-        running bump test.
+        running bump test. Note that M1M3 mirror should be in lowered
+        position.
         """
 
         for comp in self.mtcs.components_attr:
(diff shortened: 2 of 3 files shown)


python/lsst/ts/standardscripts/maintel/m1m3/check_actuators.py
```diff
@@ -86,15 +88,18 @@ async def assert_feasibility(self):
                 setattr(self.mtcs.check, comp, False)
 
         # Check all enabled, liveliness and the m1m3 detailed state
+        self.expected_states = {DetailedState.PARKED, DetailedState.PARKEDENGINEERING}
```
tribeiro, line 91: Is there a reason for this to be a class attribute? I don't think it is used anywhere else so you can drop the `self` and make it a local variable.

python/lsst/ts/standardscripts/maintel/m1m3/check_actuators.py
```diff
@@ -86,15 +88,18 @@ async def assert_feasibility(self):
...
+        self.detailed_state = await self.mtcs.assert_m1m3_detailed_state(
```
tribeiro, line 92: This method does not return the detailed state. It asserts that m1m3 is in the expected states and raises an exception if they are not.

If you want a custom error message you need to do something like this:

```
try:
    await self.mtcs.assert_m1m3_detailed_state(expected_states)
except AssertionError:
    raise RuntimeError(
        f"M1M3 detailed should be in one of the following detailed states: {detailed_states}."
        "Lower M1M3 before continuing."
)
```

I would probably do this after the other checks though. Because if m1m3 is not alive or not in enabled state this will also be false.

tribeiro (commented):
Here some initial comments.

python/lsst/ts/standardscripts/maintel/m1m3/check_actuators.py
dsanmartim: I forgot to remove it, as I planned to use it in the cleanup method before you modified the `mtcs.stop_m1m3_bump_test` method.

python/lsst/ts/standardscripts/maintel/m1m3/check_actuators.py
dsanmartim: Yes, of course :(

doc/version_history.rst
```diff
@@ -9,6 +9,7 @@ Version History
 v1.21.0
 -------
 
+* Update the ``maintel/m1m3/check_actuators.py`` script with improved logging and detailed state assertions.
```
tribeiro, line 12: Version 1.21.0 was already released. 

Please, add new version 1.21.1

python/lsst/ts/standardscripts/maintel/m1m3/check_actuators.py
```diff
@@ -77,25 +80,29 @@ def __init__(self, index, add_remotes: bool = True):
...
-                }
-            ),
         )
+        # Check if m1m3 detailed state is either PARKED or PARKEDENGINEERING
+        expected_states = {DetailedState.PARKED, DetailedState.PARKEDENGINEERING}
+        try:
+            await self.mtcs.assert_m1m3_detailed_state(expected_states)
+        except AssertionError:
+            raise RuntimeError(
+                f"M1M3 detailed should be in one of the following detailed states: {expected_states}."
```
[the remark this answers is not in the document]
pkubanek: The message shall be a variation of:
```
M1M3 needs to be parked for the bump test to proceed. Please lower the mirror, or enable M1M3 CSC.
```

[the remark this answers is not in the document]
dsanmartim: I was inclined to keep the message here just to give a more friendly message like "lower the mirror", but not sure if I got @pkubanek's comments. If the mirror is raised, will enabling the M1M3 CSC put it in a parked state? Can you clarify this, please?

The message would be then, something like:

"_Please park M1M3 before proceeding with the bump test. This can be done by lowering the mirror or enabling the M1M3 CSC._"

tribeiro (approved):
Looks good!

PR #122, opened 2023-06-08
DM-39533: Create the move_p2p SAL Script

tribeiro:


edennihy (approved):
A few minor comments to consider. 

At a minimum, I think we will also need to update base_track_target_and_take_image, base_take_stuttered, and base_take_image to be block scripts in the near future, and take a look at what we want to update in ts_externalscripts. We do not need to do that here, but could you please open a ticket or tickets for that work?

PR #123, opened 2023-06-09
DM-39620: Update track_target script to accept sexagesimal coordinates.

tribeiro:


python/lsst/ts/standardscripts/base_track_target.py
```diff
@@ -96,14 +96,20 @@ def get_schema(cls):
...
                     description: ICRS declination (deg).
-                    type: number
-                    minimum: -90
-                    maximum: 90
+                    anyOf:
+                      - type: number
+                        minimum: -90
+                        maximum: 90
+                      - type: string
+                        pattern: "[+-]?[0-9]{1,2}:[0-9]{1,2}:[0-9]*"
```
edennihy, line 112: Is this missing a leading carat "^"?

python/lsst/ts/standardscripts/base_track_target.py
```diff
@@ -96,14 +96,20 @@ def get_schema(cls):
...
                     description: ICRS right ascension (hour).
-                    type: number
-                    minimum: 0
-                    maximum: 24
+                    anyOf:
+                      - type: number
+                        minimum: 0
+                        maximum: 24
+                      - type: string
+                        pattern: "^[+-]?[0-9]{1,2}:[0-9]{1,2}:[0-9]*"
```
edennihy, line 104: No [+-] needed for ra, correct?

tests/test_maintel_track_target.py
```diff
@@ -151,6 +164,15 @@ async def test_configure_with_ra_dec(self):
             # Script can be configured with ra, dec only
             await self.configure_script(slew_icrs=dict(ra=1.0, dec=-10.0))
 
+    async def test_configure_with_ra_dec_sexagesimal(self):
+        async with self.make_script():
+            self.script._mtcs = unittest.mock.AsyncMock()
+            self.script._mtcs.start_task = utils.make_done_future()
+            # Script can be configured with ra, dec only
+            await self.configure_script(
+                slew_icrs=dict(ra="+0:00:00", dec="-10:00:00.0")
```
edennihy, line 173: This doesn't quite enforce sexagesimal since it should be "+00:00:00" for ra value, since we need two values in the hr column?

edennihy (approved):

tests/test_maintel_track_target.py
tribeiro: Not sure I understood your comment `"+0:00:00"` is a valid entry.

python/lsst/ts/standardscripts/base_track_target.py
tribeiro: Right, just the positive sign no negative.

python/lsst/ts/standardscripts/base_track_target.py
tribeiro: correct..

tests/test_maintel_track_target.py
edennihy: I mean for strict hexagesimal compliance it would need to be "+00:00:00"...but as long as it complies with slew_icrs I think we are fine.

edennihy: Actually, I think I'm wrong about needing the second leading 0 for hexagesimal compliance...feel free to ignore this request.

PR #126, opened 2023-06-20
DM-39637: Add support for generating LVV metadata in Script execution.

tribeiro:


edennihy (commented):
Would the 240s timeout limit be too short for AuxTel? That script isn't in use right now, but maybe we should just add a config overwrite for `auxtel/track_target.py`

edennihy (approved):
One minor comment to consider, otherwise approved.

PR #127, opened 2023-06-23
DM-39740: added enable/disable_m1m3_balance_system sal scripts

psferguson:
Unit tests for new scripts currently pass

### DIFF
--- a/doc/version_history.rst
+++ b/doc/version_history.rst
@@ -6,6 +6,10 @@
 Version History
 ===============
 
+v1.24.0
+-------
+* Add new ``maintel/m1m3/enable_m1m3_balance_system.py`` and ``maintel/m1m3/disable_m1m3_balance_system.py`` sal scripts and associated files.
+
 v1.23.1
 -------
 
--- a/python/lsst/ts/standardscripts/maintel/m1m3/enable_m1m3_balance_system.py
+++ b/python/lsst/ts/standardscripts/maintel/m1m3/enable_m1m3_balance_system.py
@@ -1,0 +1,72 @@
+# This file is part of ts_standardscripts
+#
+# Developed for the LSST Telescope and Site Systems.
+# This product includes software developed by the LSST Project
+# (https://www.lsst.org).
+# See the COPYRIGHT file at the top-level directory of this distribution
+# for details of code ownership.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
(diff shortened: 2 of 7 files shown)


doc/version_history.rst
```diff
@@ -6,6 +6,10 @@
 Version History
 ===============
 
+v1.23.2
```
tribeiro, line 9: When adding new scripts, please make it a new minor release instead of patch, e.g. v1.24.0

python/lsst/ts/standardscripts/maintel/m1m3/enable_m1m3_balance_system.py
```diff
@@ -0,0 +1,72 @@
...
+
+
+    """
+
+    def __init__(self, index, add_remotes: bool = True):
+        super().__init__(index=index, descr="Enable M1M3 force balance system")
+
+        mtcs_usage = None if add_remotes else MTCSUsages.DryTest
+
+        self.mtcs = MTCS(self.domain, intended_usage=mtcs_usage, log=self.log)
```
tribeiro, line 59: Check the comment I left in the other script about using `configure_tcs`

tribeiro (approved):
Looks good! thanks for the updates! 

I left a comment about the version, please, take a look before merging.

python/lsst/ts/standardscripts/maintel/m1m3/enable_m1m3_balance_system.py
psferguson: By other script, do you mean defining a configure_tcs method and calling await configure_tcs in the __init__?
I guess sort of like here: https://github.com/lsst-ts/ts_standardscripts/blob/0c6ea98ec04a251644d9d96f6b91d5d17cccefde/python/lsst/ts/standardscripts/maintel/move_p2p.py#L46-L56

tribeiro: Oh! Just ignore this comment. I wrote a larger comment before this one but honestly, I don't think it is worth the hassle. I think you van just rebase and fix the version history and we're good.

doc/version_history.rst
```diff
@@ -6,6 +6,10 @@
 Version History
 ===============
 
+v1.24.0
+-------
+* Add new ``maintel/m1m3/enable_m1m3_balance_sistem.py`` and ``maintel/m1m3/disable_m1m3_balance_sistem.py`` sal scripts and associated files.
```
pkubanek, line 11: system?

psferguson: Good catch, thanks.

psferguson:
@pkubanek brought up the point that after enabling we should ensure the hardpoints are settled before the script completes.  He suggests developing a flag to compute this state, and including a check in the script. We decided that belongs on a different ticket: DM-39816

PR #129, opened 2023-07-07
DM-39741: Update MTCS methods to handle m1m3 force balance system.

tribeiro:


python/lsst/ts/standardscripts/maintel/laser_tracker/align.py
```diff
@@ -35,8 +35,8 @@
 
 # TODO (DM-38880) - Subsitute by class in idl.enum when available
```
edennihy, line 36: This ticket is merged, can you replace these with the AlignComponent class there?

Jenkinsfile
```diff
@@ -2,6 +2,6 @@
 DevelopPipeline(
     name: "ts_standardscripts",
     module_name: "lsst.ts.standardscripts",
-    extra_packages: ["lsst-ts/ts_observatory_control"],
+    extra_packages: ["lsst-ts/ts_cRIOpy", "lsst-ts/ts_observatory_control"],
```
edennihy, line 5: Should this be "lsst-ts/ts_criopy" instead? Not sure if this is case-sensitive here...

doc/version_history.rst
```diff
@@ -6,6 +6,23 @@
...
+  * Skip alignment if tolerances are zero.
+  * Get last ``offsetPublished`` if new event is not available.
+  * Fix enum values.
+  * Skip error if laserTracker status is not available.
+
+* In ``system_wide_shutdown.py``, add more logging information.
+* Update ``tests/test_maintel_home_both_axes.py`` to check that force balance was disabled before homing.
+* In ``maintel/home_both_axes.py``, update execution to switch off force balance before homing.
+
+* Update Jenkinsfile to add ts_cRIOpy as an extra package.
```
edennihy, line 23: Also here and on the line below, package name should be updated to "ts_criopy"

edennihy (approved):
A few minor comments to consider, otherwise approved.

Jenkinsfile
tribeiro: No, this needs to be the name of the GitHub package, which is the `ts_cRIOpy` version.

doc/version_history.rst
tribeiro: yeah, the name of the package is actually `ts_cRIOpy`... 🤷

python/lsst/ts/standardscripts/maintel/laser_tracker/align.py
tribeiro: No, not yet.. only when the new cycle is released..

PR #134, opened 2023-07-24
In auxtel/daytime_checkout/latiss_checkout.py script and unit test, a…

edennihy:
…dd check to linear stage position.


tribeiro (approved):
Looks good

PR #137, opened 2023-09-12
Tickets/DM-40213

gmegh:


doc/version_history.rst
```diff
@@ -6,6 +6,12 @@
 Version History
 ===============
 
+v1.25.5
```
tribeiro, line 9: New Scripts should be minor releases.

python/lsst/ts/standardscripts/maintel/closed_loop.py
```diff
@@ -0,0 +1,519 @@
...
+    M2_B14 = 43
+    M2_B15 = 44
+    M2_B16 = 45
+    M2_B17 = 46
+    M2_B18 = 47
+    M2_B19 = 48
+    M2_B20 = 49
+
+
+class ClosedLoop(salobj.BaseScript, metaclass=abc.ABCMeta):
```
tribeiro, line 101: Not sure this is an appropriate name for the script. I think usually we want to reflect an action performed by the control system. Maybe drop the "d" and call it `CloseLoop`? However, I think the intention here is to create a base script to perform close loop with both ComCam and LSSTCam. If this is the case, I'd call this `BaseCloseLoop`.

I see that you prepared the class to do it (by adding the `metaclass`), however there's nothing in the implementation on that regard. I think you want to split this one in 2 scripts; a base and a ComCam implementation, leaving the LSSTCam implementation for later. Or maybe even doing the LSSTCam already after we merge the PR in ts_observatory_control.

python/lsst/ts/standardscripts/maintel/closed_loop.py
```diff
@@ -0,0 +1,519 @@
...
+STD_TIMEOUT = 10
+
+
+# TODO - Subsitute by class in idl.enum when available
+class ClosedLoopMode(enum.IntEnum):
+    CWFS = 0
+    FAM = 1
+
+
+class DOFComponent(enum.IntEnum):
```
tribeiro, line 48: Do you want to move this enum to ts_observatory_control as we spoke?

python/lsst/ts/standardscripts/maintel/closed_loop.py
```diff
@@ -0,0 +1,519 @@
...
+import enum
+import types
+import typing
+
+import numpy as np
+import yaml
+from lsst.ts import salobj
+from lsst.ts.observatory.control import RemoteGroup
+
+# TO-DO: Remove this import once the new LSSTCam is available
```
tribeiro, line 34: If you are making a generic implementation here, then, you don't want to import these here now. See my comment about splitting this into 2 scripts.

python/lsst/ts/standardscripts/maintel/closed_loop.py
```diff
@@ -0,0 +1,519 @@
...
+    apply_corrections attribute.
+    """
+
+    def __init__(self, index=1, add_remotes=True, descr=""):
+        super().__init__(
+            index=index,
+            descr=descr,
+        )
+
+        self.mtaos = RemoteGroup(
```
tribeiro, line 141: MTAOS is part of `MTCS`

python/lsst/ts/standardscripts/maintel/closed_loop.py
```diff
@@ -0,0 +1,519 @@
...
+            domain=self.domain,
+            components=["MTAOS"],
+            intended_usage=None if add_remotes else Usages.DryTest,
+            log=self.log,
+        )
+
+        mtcs_usage = None if add_remotes else MTCSUsages.DryTest
+        self.mtcs = MTCS(self.domain, intended_usage=mtcs_usage, log=self.log)
```
tribeiro, line 149: I think the more recent approach to this is setting values for these to `None` here and postponing the creating to the configuration stage. 

See how we did this in the  [move_p2p](https://github.com/lsst-ts/ts_standardscripts/blob/develop/python/lsst/ts/standardscripts/maintel/move_p2p.py) script.

python/lsst/ts/standardscripts/maintel/closed_loop.py
```diff
@@ -0,0 +1,519 @@
...
+        )
+
+        # The following attributes are set via the configuration:
+        self.filter = None
+        self.grating = None
+
+        # exposure time for the intra/extra images (in seconds)
+        self.exposure_time = None
+
+        # end of configurable attributes
```
tribeiro, line 166: I think you can remove this comment.

python/lsst/ts/standardscripts/maintel/closed_loop.py
```diff
@@ -0,0 +1,519 @@
...
+            properties:
+              mode:
+                description: >-
+                    Mode to use for the script. If set to "cwfs", the script will
+                    use the corner wavefront sensors, if set to "fam" the script
+                    will use the full array mode.
+                type: string
+                enum: ["cwfs", "fam"]
+                default: cwfs
+              rotation_angle:
```
tribeiro, line 185: Why do you need the user to provide this? you should get it from the system. However I am not even sure you need to get it as this information is available in the image header and the MTAOS should be able to determine this.

python/lsst/ts/standardscripts/maintel/closed_loop.py
```diff
@@ -0,0 +1,519 @@
...
+                type: number
+                default: 0.0
+                description: >-
+                    Rotator angle (deg). This is the angle between the camera
+                    and the boresight.
+              filter:
+                description: Which filter to use when taking intra/extra focal images.
+                type: string
+                default: empty_1
+              grating:
```
tribeiro, line 195: The Main telescope has no grating.

python/lsst/ts/standardscripts/maintel/closed_loop.py
```diff
@@ -0,0 +1,519 @@
...
+        self.lsstcam = ComCam(
+            self.domain,
+            intended_usage=lsstcam_usage,
+            log=self.log,
+            tcs_ready_to_take_data=self.mtcs.ready_to_take_data,
+        )
+
+        # The following attributes are set via the configuration:
+        self.filter = None
+        self.grating = None
```
tribeiro, line 161: The Main Telescope instruments have no grating

python/lsst/ts/standardscripts/maintel/closed_loop.py
```diff
@@ -0,0 +1,519 @@
...
+                default: empty_1
+              exposure_time:
+                description: The exposure time to use when taking images (sec).
+                type: number
+                default: 30.
+              dz:
+                description: De-focus to apply when acquiring the intra/extra focal images (mm).
+                type: number
+                default: 0.8
+              extra_focal_offset:
```
tribeiro, line 207: I don't think the MT has need of this

python/lsst/ts/standardscripts/maintel/closed_loop.py
```diff
@@ -0,0 +1,519 @@
...
+                description: >-
+                    The additional m2 defocus (mm) to apply for the extra-focal image to be the
+                    same size as the intra. This value is always positive to compensate for
+                    the change in magnification from moving the secondary mirror.
+                type: number
+                default: 0.0011
+              threshold:
```
tribeiro, line 214: Maybe instead if having a parameter for threshold and one for selecting the DoF, you could merge these two and make the DoF be a "dictionary" where users can specify the threshold. something like this:

```
dof:
    type: object
    description: >-
        Defines the DoF to use and their associated properties.
        Set the threshold to zero to disable a dof.
    properties:
        M2_dz:
            type: object
            properties:
                threshold:
                    type: number
        M2_dx:
            type: object
            properties:
                threshold:
                    type: number
        M2_dy:
            type: object
            properties:
                threshold:
                    type: number
        M2_rx:
            type: object
            properties:
                threshold:
                    type: number
        M2_ry:
            type: object
            properties:
                threshold:
                    type: number
        .
        .
        .
```

Now, there's a question about how to handle defaults. Maybe we should talk about this in a video conf.

python/lsst/ts/standardscripts/maintel/closed_loop.py
```diff
@@ -0,0 +1,519 @@
...
+
+            if checkpoint:
+                await self.checkpoint(
+                    f"[{i + 1}/{self.max_iter}]: CWFS loop starting..."
+                )
+
+                await self.checkpoint(f"[{i + 1}/{self.max_iter}]: Taking image...")
+
+            # Setting visit_id's to none so run_cwfs will take a new dataset.
+            if self.mode == ClosedLoopMode.CWFS:
```
tribeiro, line 432: Instead of having each option in a if/else statement like this, I'd suggest you extract each "mode" into a function, then make a dictionary with the mode as key and the function as value. Then here you just do:

```
    await self.operation_model_handlers[self.mode]()
```

where you would define `self.operation_model_handlers` in the `__init__ method as something like:

```
self.operation_model_handlers = {
    ClosedLoopMode.CWFS: self.handle_cwfs_mode,
    ClosedLoopMode.FAM: self.handle_fam_mode,
}
```

Then `self.handle_cwfs_mode` and `self.handle_fam_mode` will have the appropriate code extracted from here.

python/lsst/ts/standardscripts/maintel/closed_loop.py
```diff
@@ -0,0 +1,519 @@
...
+            else:
+                intra_image, extra_image = await self.take_intra_extra_focal_images()
+
+                intra_visit_id = int(intra_image[0])
+                extra_visit_id = int(extra_image[0])
+
+                self.mtaos.rem.mtaos.cmd_runWEP.set_start(
+                    visitId=intra_visit_id, extraId=extra_visit_id
+                )
+
```
tribeiro, line 454: Please, extract the following code into methods that explain what each step does.

tribeiro (commented):
Here is an initial round of comments.

python/lsst/ts/standardscripts/base_close_loop.py
```diff
@@ -0,0 +1,499 @@
...
+                    await self.checkpoint(
+                        f"[{i + 1}/{self.max_iter}]: Closed Loop converged."
+                    )
+
+                self.log.info("Closed Loop completed successfully!\n")
+                return
+
+            # If we haven't converged and apply_corrections is true,
+            # then we apply the corrections
+            elif self.apply_corrections:
```
tribeiro, line 453: I think you always want to apply the correction, even if you already converged. This way we don't trow away any iteration. I'd move this check to be before the convergence criteria check and make them independent `if` statements, e.g.;

```
if self.apply_corrections:
    ...

if all(abs(dof_offset) < self.threshold):
    ...
```

you don't need the `else` step.

Also, if we are not applying the correction, do we want to continue taking data? we might want to break or return in that case, e.g.;

```
if self.apply_corrections:
    ...
else:
    return
```

python/lsst/ts/standardscripts/data/scripts/maintel/closed_loop.py
```diff
@@ -0,0 +1,27 @@
...
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program. If not, see <https://www.gnu.org/licenses/>.
+
+import asyncio
+
+from lsst.ts.standardscripts.maintel import ClosedLoop
```
tribeiro, line 25: There's a `ClosedLoopComCam` and a `ClosedLoopLSSTCam` but no `ClosedLoop`. I think we need 2 of these one for each of the previously mentioned scripts.

python/lsst/ts/standardscripts/maintel/close_loop_comcam.py
```diff
@@ -0,0 +1,76 @@
...
+                intended_usage=comcam_usage,
+                log=self.log,
+                tcs_ready_to_take_data=self.mtcs.ready_to_take_data,
+            )
+            await self._comcam.start_task
+        else:
+            self.log.debug("Camera already defined, skipping.")
+
+    async def assert_mode_compatibility(self) -> None:
+        """Assert that the mode is compatible with ComCam."""
```
tribeiro, line 73: Please, document that is raises an exception.

python/lsst/ts/standardscripts/base_close_loop.py
```diff
@@ -0,0 +1,499 @@
...
+
+        # Define operation mode handler function
+        self.operation_model_handlers = {
+            ClosedLoopMode.CWFS: self.handle_cwfs_mode,
+            ClosedLoopMode.FAM: self.handle_fam_mode,
+        }
+
+    @property
+    @abc.abstractmethod
+    def camera(self):
```
tribeiro, line 94: Can you add type annotation to these method? I notice below that you did `ComCam | None` and `LSSTCam | None`. Type-checking wise this will make our life harder. We currently don't have type checking on in this repo but looking forward I would like to add support for it. 

I'd recommend you only return the camera type and raise an exception if the camera is `None`.  Something like this:

```
@property
@abc.abstractmethod
def camera(self) -> BaseCamera:
    raise NotImplementedError()
```

Then for ComCam and LSSTCam you would do something like this:

```
@property
def camera(self) -> BaseCamera:
    if self._comcam is not None:
        return self._comcam
    else:
        raise RuntimeError("Camera not defined.")
```

You can either use `BaseCamera` or `ComCam` as the return type in this case.

python/lsst/ts/standardscripts/maintel/close_loop_comcam.py
```diff
@@ -0,0 +1,76 @@
...
+
+    def __init__(self, index, add_remotes=True, descr="") -> None:
+        super().__init__(index=index, add_remotes=add_remotes, descr=descr)
+
+        self.config = None
+
+        self._comcam = None
+
+    @property
+    def camera(self) -> ComCam | None:
```
tribeiro, line 54: See my comment above about the type and how to handle this

python/lsst/ts/standardscripts/maintel/close_loop_lsstcam.py
```diff
@@ -0,0 +1,69 @@
...
+
+    def __init__(self, index, add_remotes=True, descr="") -> None:
+        super().__init__(index=index, add_remotes=add_remotes, descr=descr)
+
+        self.config = None
+
+        self._lsstcam = None
+
+    @property
+    def camera(self) -> LSSTCam | None:
```
tribeiro, line 53: See my comment above about the type

tests/test_base_close_loop.py
```diff
@@ -0,0 +1,143 @@
...
+# along with this program. If not, see <https://www.gnu.org/licenses/>.
+
+import asyncio
+import random
+import unittest
+
+import numpy as np
+from lsst.ts import standardscripts
+from lsst.ts.observatory.control.utils.enums import ClosedLoopMode
+from lsst.ts.standardscripts import BaseCloseLoop
```
tribeiro, line 29: `BaseCloseLoop` is an abstract class, I don't think you can unit test it. You will have to test the concrete implementations.

tribeiro (commented):
Here some additional comments.

python/lsst/ts/standardscripts/base_close_loop.py
```diff
@@ -0,0 +1,499 @@
...
+                    f"[{i + 1}/{self.max_iter}]: Closed Loop loop starting..."
+                )
+
+                await self.checkpoint(f"[{i + 1}/{self.max_iter}]: Taking image...")
+
+            # Run the operational mode handler function.
+            await self.operation_model_handlers[self.mode]()
+
+            # Save the wavefront error
+            wavefront_error = await self.mtcs.mtaos.evt_wavefrontError.next(
```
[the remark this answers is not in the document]
gmegh: do you mean by setting flus=True?

python/lsst/ts/standardscripts/base_close_loop.py
gmegh: I think we do want to keep taking data if we are not applying the correction. The reason being that in that way we can see how the closed loop corrections evolve without applying them. We might be repositioning the telescope or checking the LUT while doing this

tests/test_base_close_loop.py
gmegh: Cool, do I need to add two unit tests? lsstcam and comcam are basically the same, is it okay if I just have an lsstcam unit test?

python/lsst/ts/standardscripts/base_close_loop.py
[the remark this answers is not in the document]
tribeiro: Sorry, no, I mean call `self.mtcs.mtaos.evt_wavefrontError.flush()` before running `await self.operation_model_handlers[self.mode]()`

tests/test_base_close_loop.py
tribeiro: one one is fine.

python/lsst/ts/standardscripts/base_close_loop.py
tribeiro: Ok, that's a good point. In this case, please, add a note in the `max_iter` description explaining that, if `apply_corrections` is `False` it will take this many pairs of images.

python/lsst/ts/standardscripts/base_close_loop.py
```diff
@@ -0,0 +1,499 @@
...
+    - "[N/MAX_ITER]: Closed converged.": Once Closed Loop reaches convergence.
+    - "[N/MAX_ITER]: Closed applying correction.": Just before
+        corrections are applied.
+
+    **Details**
+
+    This script is used to perform measurements of the wavefront error, then
+    propose dof offsets based on ts_ofc. The offsets are not applied
+    automatically and must be turned on by the user through
+    apply_corrections attribute.
```
tribeiro, line 68: Please, add a note here that if `apply_corrections` is off, the script will take a series of intra/extra focal data instead, and the number of pairs is the number of maximum iterations.

doc/version_history.rst
tribeiro, line 1: Please, make sure you reset all changes to this file.

doc/news/DM-40213.feature.rst
```diff
@@ -0,0 +1,3 @@
+Add new base_close_loop.py script, and executable.
```
tribeiro, line 1: Please, explaining what this script does.

python/lsst/ts/standardscripts/base_close_loop.py
```diff
@@ -0,0 +1,492 @@
...
+                oneOf:
+                  - type: array
+                    items:
+                      type: integer
+                      minimum: 0
+                      maximum: 49
+                  - type: array
+                    items:
+                      type: string
+                      enum: [
```
tribeiro, line 194: I think you could dynamically generate this by doing something like:

```
enum: {[dof_name.name for dof_name in DOFName]}
```

and covert this to an f-string, by adding a "f" at the top, e.g.:

```
schema_yaml = f"""
...          
```

tests/test_maintel_close_loop_lsstcam.py
```diff
@@ -0,0 +1,147 @@
...
+                filter="r",
+                used_dofs=[0, 1, 2, 3, 4],
+            )
+
+            # Run the script
+            await self.run_script()
+
+            assert all(self.state_0 == np.zeros(50))
+
+    async def test_executable(self):
```
tribeiro, line 134: Please, split this test in 2, e.g.;

```
    async def test_executable_close_loop_lsstcam(self):
        ...

    async def test_executable_close_loop_comcam(self):
        ...
```

tribeiro (commented):
Here some additional comments. I think once you address those we can probably close this one. 

you may want to run

```
git checkout develop -- doc/version_history.rst
```

to make sure all your changes to this file are reset. Also, please, expand a bit the news fragment.

python/lsst/ts/standardscripts/base_close_loop.py
tribeiro, line 1: I think this module can be inside `maintel` since it is maintel specific.

tests/test_maintel_close_loop_lsstcam.py
```diff
@@ -0,0 +1,148 @@
...
+from lsst.ts.standardscripts.maintel import CloseLoopLSSTCam
+
+random.seed(47)  # for set_random_lsst_dds_partition_prefix
+
+
+class TestCloseLoopLSSTCam(
+    standardscripts.BaseScriptTestCase, unittest.IsolatedAsyncioTestCase
+):
+    async def basic_make_script(self, index):
+        self.script = CloseLoopLSSTCam(index=index, add_remotes=False)
```
tribeiro, line 38: Instead of using `add_remotes`, set the  value of the camera and TCS classes here with a `DryTest` usage, e.g.;

```
self.script = CloseLoopLSSTCam(index=index)
self.script.mtcs = MTCS(
                domain=self.script.domain,
                intended_usage=MTCSUsages.DryTest,
                log=self.script.log,
            )
self.script._lsstcam = LSSTCam(
                domain=self.script.domain,
                intended_usage=LSSTCam.DryTest,
                log=self.script.log,
            )
```

tribeiro (commented):
Looking good. Here some additional comments!

tribeiro (approved):
Looks good! Thanks!

PR #141, opened 2023-09-26
Integrate towncrier tool for changelog management

dsanmartim:
Hi Tiago,

I just implemented `towncrier` here. It basically reflects the same config already created in the ts_externalscripts. If you have further suggestions, let me know. 

Tks!


tribeiro (approved):
Looks good...

PR #143, opened 2023-10-02
Tickets/dm 40209

edennihy:


python/lsst/ts/standardscripts/maintel/offset_camera_hexapod.py
```diff
@@ -0,0 +1,132 @@
...
+              z:
+                type: number
+                description: Offset hexapod in z axis (mm).
+              u:
+                type: number
+                description: Rx offset (deg).
+              v:
+                type: number
+                description: Ry offset (deg).
+              w:
```
tribeiro, line 81: I would remove w from the list of options. This is basically a rotation and we should never do that with the hexapod.

tribeiro (approved):
Looks good,  however, take a look at my inline comment. I think you should remove "w" from the list of axis. "w" is a rotation and we should not be doing that with the hexapod.

python/lsst/ts/standardscripts/maintel/offset_camera_hexapod.py
edennihy: Request adopted.

PR #146, opened 2023-10-12
Add new ``move_rotator.py`` SAL Script 

b1quint:
Add the new ``move_rotator.py`` SAL Script. This depends on [lsst-ts/ts_observatory_control#107](https://github.com/lsst-ts/ts_observatory_control/pull/107).

### DIFF
--- a/python/lsst/ts/standardscripts/maintel/mtrotator/move_rotator.py
+++ b/python/lsst/ts/standardscripts/maintel/mtrotator/move_rotator.py
@@ -1,0 +1,141 @@
+# This file is part of ts_standardscripts
+#
+# Developed for the LSST Telescope and Site Systems.
+# This product includes software developed by the LSST Project
+# (https://www.lsst.org).
+# See the COPYRIGHT file at the top-level directory of this distribution
+# for details of code ownership.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program. If not, see <https://www.gnu.org/licenses/>.
+
+__all__ = ["MoveRotator"]
+
+import yaml
+from lsst.ts.observatory.control.maintel.mtcs import MTCS
+
+from ...base_block_script import BaseBlockScript
+
+
+class MoveRotator(BaseBlockScript):
+    """Move the rotator to a given angle. It has the option of completing the
+    script before the rotator reaches the desired angle.
+
+    Parameters
+    ----------
+    index : `int`
+        Index of Script SAL component.
+
+    Notes
+    -----
+    **Checkpoints**
+
+    - "Start moving rotator to {angle} degrees.": Start moving rotator.
+    - "Stop script and keep rotator moving.": Stop script.
+    - "Rotator reached {angle} degrees.": Rotator reached angle.
+
+    """
+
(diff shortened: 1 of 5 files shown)


python/lsst/ts/standardscripts/maintel/mtrotator/move_rotator.py
```diff
@@ -0,0 +1,126 @@
...
+    - "Start moving rotator to {angle} degrees.": Start moving rotator.
+    - "Stop script and keep rotator moving.": Stop script.
+    - "Rotator reached {angle} degrees.": Rotator reached angle.
+
+    """
+
+    def __init__(self, index: int, add_remotes: bool = True) -> None:
+        super().__init__(index=index, descr="Move Rotator")
+
+        self.mtcs = MTCS(
```
tribeiro, line 52: Recently I have found that a better design is to postpone creating the TCS class until the `configure` stage. This allows us to intervene during the unit test and skip adding the `add_remotes` option. It also makes the Script start much quicker since the heavy burden of creating the TCS class is done after the script start. This is what I would suggest.

Here in the init you would do:

```
        self.mtcs = None

        self.rotator_velocity = 3.5  # degrees per second
        self.short_timeout = 10  # seconds
        self.long_timeout = 120  # seconds
```

Then, in the `configure` method you would include the following:

```
if self.mtcs is None:
    self.mtcs = MTCS(
        domain=self.domain,
        log=self.log,
    )

for comp in self.mtcs.components_attr:
    if comp not in ["mtrotator", "mtmount"]:
        setattr(self.mtcs.check, comp, False)

await self.mtcs.start_task
```

Then in the unit test you can probably do something like this:

```
        self.script = MoveRotator(index=index, add_remotes=False)
        self.script.mtcs = MTCS(domain=self.script.domain, intended_usage=MTCSUsages.DryTest,
            log=self.script.log)
        self.start_angle = 0.0  # degrees
        self.very_short_sleep = 0.1  # seconds
        self.script.mtcs.move_rotator = unittest.mock.AsyncMock()
```

Honestly, I wouldn't go through the trouble of checking the behavior of `move_rotator` because that is already done in ts_observatory_control. I think you can check that the `move_rotator` command is awaited with the intended parameters though.

tribeiro (commented):
Please, see my inline comment.

python/lsst/ts/standardscripts/maintel/mtrotator/move_rotator.py
```diff
@@ -0,0 +1,142 @@
...
+    -----
+    **Checkpoints**
+
+    - "Start moving rotator to {angle} degrees.": Start moving rotator.
+    - "Stop script and keep rotator moving.": Stop script.
+    - "Rotator reached {angle} degrees.": Rotator reached angle.
+
+    """
+
+    def __init__(self, index: int, add_remotes: bool = True) -> None:
```
tribeiro, line 49: I think you can remove the `add_remotes` parameter now.

python/lsst/ts/standardscripts/maintel/mtrotator/move_rotator.py
```diff
@@ -0,0 +1,142 @@
...
+        """Set the metadata for the script."""
+        metadata.duration = self.long_timeout
+
+    async def run_block(self):
+        """Run the script."""
+        await self.checkpoint(f"Start moving rotator to {self.target_angle} degrees.")
+        await self.mtcs.move_rotator(
+            angle=self.target_angle, wait_for_complete=self.wait_for_complete
+        )
+        if self.wait_for_complete:
```
tribeiro, line 139: Not sure I understand why you included this checkpoints here. Is the idea to provide a way for people to pause the script? I suggest you change it to be a single message, regardless of the value of `self.wait_for_complete`. Maybe something like:

```
await self.checkpoint(f"Move rotator finished.")
```

Or, if you want to provide information about the `wait_for_complete`, something like this:

```
await self.checkpoint(f"Move rotator finished; wait for complete: {self.wait_for_complete}.")
```

tribeiro (approved):
Looks good.. Take a look at my comments before merging please!

PR #147, opened 2023-10-12
Tickets/dm 41083 - Add maintel SAL script to stop_rotator 

edennihy:


python/lsst/ts/standardscripts/maintel/stop_rotator.py
```diff
@@ -0,0 +1,65 @@
...
+    Stopping rotator: Before issuing stop command.
+    Done: After issuing stop command.
+    """
+
+    def __init__(self, index, add_remotes: bool = True):
+        super().__init__(index=index, descr="MTCS stop rotator.")
+
+        mtcs_usage = None if add_remotes else MTCSUsages.DryTest
+
+        self.mtcs = MTCS(domain=self.domain, intended_usage=mtcs_usage, log=self.log)
```
tribeiro, line 49: I would suggest moving the creating of the MTCS object into the `configure` method. So, here you would do:

```
self.mtcs = None
```

Then, in the `configure` method you would do:

```
if self.mtcs is None:
    self.mtcs = MTCS(domain=self.domain, log=self.log)
```

Then remove the `add_remotes` from the class initialization and the `mtcs_usage` above. 


In the unit test you can do something like this:

```
        self.script = StopRotator(index=index)

        self.script.mtcs = unittest.mock.AsyncMock()
        self.script.mtcs.configure_mock(tel_settle_time=3.0)  # need this for the set_metadata method.
        return (self.script,)
```

Then, you make a unit test like this:

```
    async def test_run(self):
            await self.configure_script()

            await self.run_script()

            self.script.mtcs.stop_rotator.assert_awaited_once()
```

tribeiro (commented):
Please, take a look at my inline comment.

edennihy:
Recommended changes adopted.

python/lsst/ts/standardscripts/maintel/stop_rotator.py
```diff
@@ -0,0 +1,63 @@
...
+        self.mtcs = None
+
+    @classmethod
+    def get_schema(cls):
+        return None
+
+    async def configure(self, config):
+        if self.mtcs is None:
+            self.mtcs = MTCS(domain=self.domain, log=self.log)
+        pass
```
tribeiro, line 55: you can remove this `pass` here.

tribeiro (approved):
Alright! looks good! please, check my final comment and reorganize the commits before merging.

PR #164, opened 2023-12-19
Tickets/DM-41611

dsanmartim:
Hi Tiago, 

Here is the script to enable/disable M2 closed-loop.

Tks!


tribeiro (approved):
Looks good! thanks!

PR #168, opened 2024-02-05
Make SAL Script to mute watcher alarm

b1quint:
Create a SAL Script with the same inputs as `Watcher.cmd_mute`.

### DIFF
--- a/python/lsst/ts/standardscripts/mute_alarms.py
+++ b/python/lsst/ts/standardscripts/mute_alarms.py
@@ -1,0 +1,117 @@
+# This file is part of ts_standardscripts
+#
+# Developed for the LSST Telescope and Site Systems.
+# This product includes software developed by the LSST Project
+# (https://www.lsst.org).
+# See the COPYRIGHT file at the top-level directory of this distribution
+# for details of code ownership.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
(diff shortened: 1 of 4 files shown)


tests/test_mute_alarms.py
```diff
@@ -0,0 +1,69 @@
...
+            await self.configure_script(
+                name=".*",
+                mutedBy="test",
+                duration=10,
+                severity="WARNING",
+            )
+
+            # Run the script
+            await self.run_script()
+
```
tribeiro, line 61: Here make an assertion that the `mute` command was awaited with the expected set of parameters:

```
self.script.watcher.cmd_mute.set_start.assert_awaited_with(.....)
```

python/lsst/ts/standardscripts/mute_alarms.py
```diff
@@ -0,0 +1,118 @@
...
+        """Run the script."""
+        self.log.info(f"Muting alarm(s) {self.name} for {self.duration} seconds.")
+        await self.watcher.cmd_mute.set_start(
+            name=self.name,
+            duration=self.duration,
+            severity=self.severity,
+            mutedBy=self.mutedBy,
+            timeout=self.std_timeout,
+        )
+        self.log.info(f"Unmuting alarm(s) {self.name}.")
```
tribeiro, line 118: Why is it saying "unmuting alarm"?

tribeiro (approved):
Looks great! thanks! However, there's a log message in the run method that probably needs rephrasing.. it says "Unmuting alarm(s) ...".

I think you can probably remove that log.

PR #172, opened 2024-02-19
Tickets/dm 42516

dsanmartim:
This is a draft PR, with the `run_block` unit test still failing. I need some light on what is happening :( 

Tks!


tribeiro (commented):
One general comment about how you structured things here, you have objects for each individual camera at the Script level and then you also define `self.camera_setups`, which is a list of `CameraSetup`, that also holds a reference to the camera. I think you could refactor the code to remove the bare references (e.g. `self.comcam`, `self.lsstcam`, `self.gen_cams`) and only keep `self.camera_setups`.

Also, I think it might be best to use a dictionary instead of a list so you can keep named references to these objects.. Also, make sure you initialize `self.camera_setups` in the `__init__` method.

tests/test_maintel_take_image_anycam.py
```diff
@@ -0,0 +1,403 @@
...
+
+class TestTakeImageAnyCam(
+    standardscripts.BaseScriptTestCase, unittest.IsolatedAsyncioTestCase
+):
+    async def basic_make_script(self, index):
+        self.script = TakeImageAnyCam(index=index)
+
+        # Mock MTCS instances and its methods
+        self.script.mtcs = unittest.mock.AsyncMock()
+        self.script.mtcs.start_task = utils.make_done_future()
```
tribeiro, line 39: You can remove this. It would only be called in the `configure` method if `mtcs` is not defined.

tests/test_maintel_take_image_anycam.py
```diff
@@ -0,0 +1,403 @@
...
+
+        self.script.lsstcam = unittest.mock.AsyncMock()
+        self.script.lsstcam.start_task = utils.make_done_future()
```
tribeiro, line 42: You can remove this.

tests/test_maintel_take_image_anycam.py
```diff
@@ -0,0 +1,403 @@
...
+
+        self.script.comcam = unittest.mock.AsyncMock()
+        self.script.comcam.start_task = utils.make_done_future()
```
tribeiro, line 45: You can remove this.

tests/test_maintel_take_image_anycam.py
```diff
@@ -0,0 +1,403 @@
...
+
+        self.script.generic_cams = []
+        # Mock specific behaviors for each Generic Camera
+        for i in [101, 102, 103]:  # Adjust indexes as needed
+            gencam = unittest.mock.MagicMock(name=f"GenericCam_{i}")
+            gencam.start_task = utils.make_done_future()
```
tribeiro, line 51: You can remove this.

tests/test_maintel_take_image_anycam.py
```diff
@@ -0,0 +1,403 @@
...
+            gencam.name = f"GenericCam_{i}"
+            self.script.generic_cams.append(gencam)
+
+        return (self.script,)
+
+    @contextlib.asynccontextmanager
+    async def make_dry_script(self):
```
tribeiro, line 58: I don't think you need this, just use the `make_script` instead.

tests/test_maintel_take_image_anycam.py
```diff
@@ -0,0 +1,403 @@
...
+        async with self.make_script():
+            # Use mock_cameras to setup all camera
+            self.mock_cameras()
+
+            # Mock the take_images_with_camera method
+            self.script.take_images_with_camera = unittest.mock.AsyncMock()
+
+            yield
+
+    async def mock_cameras(self):
```
tribeiro, line 69: You don't need this. The mock you do in `basic_make_script` should be enough.

tests/test_maintel_take_image_anycam.py
```diff
@@ -0,0 +1,403 @@
...
+            gencam.assert_liveliness = unittest.mock.AsyncMock()
+            gencam.assert_all_enabled = unittest.mock.AsyncMock()
+            gencam.take_imgtype = unittest.mock.AsyncMock()
+            gencam.read_out_time = 1
+            gencam.shutter_time = 0.1
+            self.script.generic_cams.append(gencam)
+
+            yield
+
+    async def configure_and_mock_camera_setups(self, **config):
```
tribeiro, line 101: I don't think you need this. You are already creating the mocks in the `basic_make_script` so just call `configure_script`

python/lsst/ts/standardscripts/maintel/take_image_anycam.py
```diff
@@ -0,0 +1,448 @@
...
+            self.log.debug("Creating MTCS.")
+            self.mtcs = MTCS(domain=self.domain, log=self.log)
+            await self.mtcs.start_task
+        else:
+            self.log.debug("MTCS already defined, skipping.")
+
+        self.camera_setups = []
+
+        # Initialize LSSTCam if configured
+        if self.lsstcam is None and hasattr(config, "lsstcam"):
```
tribeiro, line 271: If you adopt my suggestion to remove these references and only hold a dict of `CameraSetup`, you could do something like this instead:

```
if "lsstcam" not in self.camera_setups and hasattr(config, "lsstcam"):
    self.log.debug("Creating LSSTCam.")
    lsstcam = LSSTCam(self.domain, log=self.log)
    await lsstcam.start_task
    self.camera_setups["lsstcam"] = CameraSetup(self.lsstcam, config.lsstcam, "LSSTCam")
```

Then do the same with "comcam" and "gencam".

python/lsst/ts/standardscripts/maintel/take_image_anycam.py
```diff
@@ -0,0 +1,448 @@
...
+        if len(self.generic_cams) == 0 and hasattr(config, "gencam"):
+            self.log.debug("Creating Generic Cameras.")
+
+            async def init_gencam(gencam_config):
+                gencam = GenericCamera(
+                    gencam_config["index"], domain=self.domain, log=self.log
+                )
+                await gencam.start_task
+                # Store the camera instance directly in self.generic_cams
+                self.generic_cams.append(gencam)
```
tribeiro, line 300: if you adopt my suggestion to remove `self.lsstcam`, `self.comcam` and `self.generic_cams`, you can remove this.

tests/test_maintel_take_image_anycam.py
```diff
@@ -0,0 +1,403 @@
...
+            gencam = unittest.mock.MagicMock(name=f"GenericCam_{i}")
+            gencam.start_task = utils.make_done_future()
+            gencam.assert_liveliness = unittest.mock.AsyncMock()
+            gencam.assert_all_enabled = unittest.mock.AsyncMock()
+            gencam.take_imgtype = unittest.mock.AsyncMock()
+            gencam.read_out_time = 1
+            gencam.shutter_time = 0.1
+            self.script.generic_cams.append(gencam)
+
+            yield
```
tribeiro, line 99: are you sure you need this yield here?

tribeiro (commented):
Here some additional comments.

tests/test_maintel_take_image_anycam.py
dsanmartim: No. The mMethod was removed anyway.

tribeiro (approved):
Looks good! but please remove the commented code before merging...

PR #176, opened 2024-03-14
Tickets/dm 43298

edennihy:


python/lsst/ts/standardscripts/base_take_image.py
```diff
@@ -153,6 +195,30 @@ def set_metadata(self, metadata):
             * nimages
         )
 
+    def set_additional_nextvisit_metadata(self, metadata):
+        metadata.nimages = len(self.config.exp_times)
+        metadata.coordinateSystem = MetadataCoordSys.ICRS
+        if (self.ra is not None) and (self.dec is not None):
```
tribeiro, line 201: Pleas, see my suggestion above about removing `self.ra` and `self.dec`, they are not really doing anything as the configuration does not have these fields.

python/lsst/ts/standardscripts/base_take_image.py
```diff
@@ -153,6 +195,30 @@ def set_metadata(self, metadata):
...
+            radec_icrs = ICRS(
+                Angle(self.nextVisit_metadata["ra"], unit=astropy.units.hourangle),
```
tribeiro, line 203: As I mentioned above, the metadata session is already part of `self.config`, so instead of duplicating this information just use it from it, e.g.:

```
Angle(self.config.nextVisit_metadata["ra"], unit=astropy.units.hourangle)
```

but also, please, rename this to either `next_visit_metadata` or `visit_metadata` or, preferably, just `metadata`.

python/lsst/ts/standardscripts/base_take_image.py
```diff
@@ -153,6 +195,30 @@ def set_metadata(self, metadata):
...
+                Angle(self.nextVisit_metadata["dec"], unit=astropy.units.deg),
```
tribeiro, line 204: Same comment as above.

python/lsst/ts/standardscripts/base_take_image.py
```diff
@@ -153,6 +195,30 @@ def set_metadata(self, metadata):
...
+            )
+            metadata.position = [radec_icrs.ra.deg, radec_icrs.dec.deg]
+        if self.rot_sky is not None:
+            metadata.rotationSystem = MetadataRotSys.SKY
+            metadata.cameraAngle = self.nextVisit_metadata["rot_sky"]
```
tribeiro, line 209: Same comment as above.

python/lsst/ts/standardscripts/base_take_image.py
```diff
@@ -153,6 +195,30 @@ def set_metadata(self, metadata):
...
+            metadata.instrument = self.get_instrument_name()
+            if self.program is not None:
```
tribeiro, line 211: The Script does not have this field so you can remove this.

python/lsst/ts/standardscripts/base_take_image.py
```diff
@@ -153,6 +195,30 @@ def set_metadata(self, metadata):
...
+                metadata.survey = self.config.program
+            instrument_configuration = self.get_instrument_configuration()
```
tribeiro, line 213: Maybe instead of doing this here, that tests for the `grating` thing, make a new method that just returns the instrument configuration label. Maybe something like `get_instrument_configuration_label`? You could actually have a default implementation that just returns`instrument_configuration["filter"]` and then overwrite it in `LATISS` to return `f"{filt}~{grating}"`.

edennihy:
Thanks, all recommended changes were incorporated.

PR #183, opened 2024-04-18
Tickets/dm 41806

dsanmartim:
Hi Tiago.

It took me some time to figure out how to interact with the ESS (not sure if used it correctly though, should I have used an index?) to get wind speed, but hopefully, this version will not need major changes. 

In the last commit, you will find a more "robust" way to determine if it is safe to open. In the first one, I was just doing a pretty straightforward check of the median wind speed. 

Also, I noted that unit test ``test_run_script_cannot_determine_wind_speed`` seems to be deselected, not giving any output status. Do you know why that happens?

### DIFF
--- a/python/lsst/ts/standardscripts/auxtel/atdome/open_dropout_door.py
+++ b/python/lsst/ts/standardscripts/auxtel/atdome/open_dropout_door.py
@@ -1,0 +1,166 @@
+# This file is part of ts_standardscripts
+#
+# Developed for the LSST Telescope and Site Systems.
+# This product includes software developed by the LSST Project
+# (https://www.lsst.org).
+# See the COPYRIGHT file at the top-level directory of this distribution
+# for details of code ownership.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program. If not, see <https://www.gnu.org/licenses/>.
+
+__all__ = ["OpenDropoutDoor"]
+
+import asyncio
+
+from lsst.ts import salobj
+from lsst.ts.observatory.control.auxtel.atcs import ATCS
+
+STD_TIMEOUT = 5
+
+
+class OpenDropoutDoor(salobj.BaseScript):
+    """
+    A SAL script for opening the dropout door of the Auxiliary Telescope's
+    dome based on current wind conditions as reported by the Environmental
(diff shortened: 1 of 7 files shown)


tests/test_auxtel_open_dropout_door.py
```diff
@@ -0,0 +1,131 @@
...
+
+
+class TestOpenDropoutDoor(
+    standardscripts.BaseScriptTestCase, unittest.IsolatedAsyncioTestCase
+):
+
+    async def basic_make_script(self, index):
+        # Create an instance of the script.
+        self.script = OpenDropoutDoor(index=index)
+        self.atcs_mock = ATCSMock()
```
tribeiro, line 42: Avoid using the `ATCSMock` these are going to be deprecated, it is better to mock the behavior using Mock.

tests/test_auxtel_open_dropout_door.py
```diff
@@ -0,0 +1,131 @@
...
+                # Make sure it was not awaited
+                self.script.atcs.open_dropout_door.assert_awaited_once()
+
+    async def test_executable(self):
+        scripts_dir = standardscripts.get_scripts_dir()
+        script_path = scripts_dir / "auxtel" / "atdome" / "open_dropout_door.py"
+        await self.check_executable(script_path)
+
+
+if __name__ == "__main__":
```
tribeiro, line 130: You can remove this

tests/test_auxtel_close_dropout_door.py
```diff
@@ -0,0 +1,56 @@
...
+from lsst.ts.observatory.control.mock import ATCSMock
+from lsst.ts.standardscripts.auxtel.atdome import CloseDropoutDoor
+
+
+class TestCloseDropoutDoor(
+    standardscripts.BaseScriptTestCase, unittest.IsolatedAsyncioTestCase
+):
+    async def basic_make_script(self, index):
+        self.script = CloseDropoutDoor(index=index)
+        self.atcs_mock = ATCSMock()
```
tribeiro, line 34: It is better not to use `ATCSMock`, they are to be deprecated, use mocks instead.

tribeiro (commented):
Here some initial comments.

python/lsst/ts/standardscripts/auxtel/atdome/open_dropout_door.py
```diff
@@ -0,0 +1,165 @@
...
+                timeout=STD_TIMEOUT,
+            )
+
+            if air_flow.speedStdDev > std_dev_threshold:
+                # Increase sensitivity to wind when variability is high
+                median_threshold *= 0.8
+                max_threshold *= 0.8
+
+            if air_flow.speed > median_threshold or air_flow.maxSpeed > max_threshold:
+                raise RuntimeError(
```
tribeiro, line 141: So, this method is called `assert_` but you are raising a `RuntimeError`. I think it would be best to raise an `AssertionError`. You could do an assert here instead of if/raise.

Something like this: 

```
assert air_flow.speed < median_threshold and air_flow.maxSpeed < max_threshold, "...."
```

tribeiro (approved):
Looks good! I only have a small comment on how you are raising the exception on the assert method. I think it would be worth changing it, but I leave that to you.

PR #187, opened 2024-05-14
Add take_triplet_comcam script and unit tests

gmegh:


python/lsst/ts/standardscripts/maintel/take_triplet_comcam.py
```diff
@@ -0,0 +1,244 @@
...
+    def __init__(self, index, descr="Take triplet with ComCam.") -> None:
+        super().__init__(index=index, descr=descr)
+
+        self.config = None
+        self.mtcs = None
+        self._comcam = None
+
+    @property
+    def camera(self) -> ComCam:
+        return self._comcam
```
tribeiro, line 59: Is there really any need to do this? Why not just have `self.camera` as an attribute? Unless you are thinking about also having this script work for `LSSTCam`, in which case, you might want to do that already.

python/lsst/ts/standardscripts/maintel/take_triplet_comcam.py
```diff
@@ -0,0 +1,244 @@
...
+                default: 30.
+              dz:
+                description: De-focus to apply when acquiring the intra/extra focal images (microns).
+                type: number
+                default: 1500.
+              n_iterations:
+                description: Number of triplets.
+                type: integer
+                default: 1
+              program:
```
tribeiro, line 90: Program and reason only makes sense if this is a subclass of `BaseBlockScript`.

python/lsst/ts/standardscripts/maintel/take_triplet_comcam.py
```diff
@@ -0,0 +1,244 @@
...
+        # Set exposure time
+        self.exposure_time = config.exposure_time
+
+        # Set intra/extra focal offsets
+        self.dz = config.dz
+
+        # Set maximum number of iterations
+        self.n_iterations = config.n_iterations
+
+        # Set program and reason
```
tribeiro, line 129: Again, only makes sense if this is a subclass of `BaseBlockScript`

python/lsst/ts/standardscripts/maintel/take_triplet_comcam.py
```diff
@@ -0,0 +1,244 @@
...
+
+        Parameters
+        ----------
+        metadata : `salobj.type_hints.BaseMsgType`
+            Script metadata topic. The information is set on the topic
+            directly.
+        """
+        # Estimated duration is maximum number of iterations multiplied by
+        # 3 (data triplet) multiplied by the time it takes to take an image
+        # plus estimation on processing the data (10s)
```
tribeiro, line 144: Processing the data? or you mean reading out? Is there any processing going on in the script?

python/lsst/ts/standardscripts/maintel/take_triplet_comcam.py
```diff
@@ -0,0 +1,244 @@
...
+        else:
+            self.log.debug("MTCS already defined, skipping.")
+
+    async def take_triplet(
+        self,
+    ) -> None:
+        """Take triplet (intra focal, extra focal and in-focus)
+        image sequence.
+        """
+        # Take intra focal image
```
tribeiro, line 187: With the log message below this comment is moot.

python/lsst/ts/standardscripts/maintel/take_triplet_comcam.py
```diff
@@ -0,0 +1,244 @@
...
+
+        self.log.debug("Moving to in-focus position")
+
+        # Move the hexapod back to in focus position
+        await self.mtcs.move_camera_hexapod(x=0, y=0, z=self.dz, u=0, v=0)
+
+        self.log.info("Taking in-focus image")
+
+        # Take in-focus iamge
+        await self.camera.take_cwfs(
```
tribeiro, line 230: in-focus image should be `take_acq`... The `take_cwfs` is reserved for the intra/extra images.

python/lsst/ts/standardscripts/maintel/take_triplet_comcam.py
```diff
@@ -0,0 +1,244 @@
...
+            exptime=self.exposure_time,
+            n=1,
+            group_id=self.group_id,
+            filter=self.filter,
+            reason="INFOCUS" + ("" if self.reason is None else f"_{self.reason}"),
+            program=self.program,
+        )
+
+    async def run(self) -> None:
```
tribeiro, line 239: You may have to change this is you accept my suggestion to use a `BaseBlockScript` instead.

tests/test_maintel_take_triplet_comcam.py
```diff
@@ -0,0 +1,100 @@
...
+
+import random
+import unittest
+
+from lsst.ts import standardscripts
+from lsst.ts.observatory.control.maintel.comcam import ComCam, ComCamUsages
+from lsst.ts.observatory.control.maintel.mtcs import MTCS, MTCSUsages
+from lsst.ts.standardscripts.maintel import TakeTripletComCam
+
+random.seed(47)  # for set_random_lsst_dds_partition_prefix
```
tribeiro, line 30: I don't think you need this anymore.

python/lsst/ts/standardscripts/maintel/take_triplet_comcam.py
gmegh: No processing going on. Changed to reading out.

python/lsst/ts/standardscripts/maintel/take_triplet_comcam.py
```diff
@@ -0,0 +1,248 @@
...
+                + self.camera.shutter_time
+            )
+        )
+        metadata.filter = f"{self.filter}"
+
+    async def assert_feasibility(self) -> None:
+        """Verify that the telescope and camera are in a feasible state to
+        execute the script.
+        """
+        await self.mtcs.assert_all_enabled()
```
tribeiro, line 158: You can actually do this concurrently with: 

```
await asyncio.gather(
    self.mtcs.assert_all_enabled(),
    self.camera.assert_all_enabled()
)
```

python/lsst/ts/standardscripts/maintel/take_triplet_comcam.py
```diff
@@ -0,0 +1,248 @@
...
+
+    async def take_triplet(
+        self,
+    ) -> None:
+        """Take triplet (intra focal, extra focal and in-focus)
+        image sequence.
+        """
+        self.log.debug("Moving to intra-focal position")
+
+        await self.mtcs.move_camera_hexapod(x=0, y=0, z=self.dz, u=0, v=0)
```
tribeiro, line 193: I just realized that you are using the move command and not offset. 

This will cause the hexapod to move to a position that is +dz of the LUT value, regardless of the current position of the hexapod. Is this really what you want to do? it is also going to reset any offset in x, y, u and v.

I think you want to use `offset_camera_hexapod` instead.

python/lsst/ts/standardscripts/maintel/take_triplet_comcam.py
```diff
@@ -0,0 +1,248 @@
...
+            reason="INTRA" + ("" if self.reason is None else f"_{self.reason}"),
+            program=self.program,
+        )
+
+        self.log.debug("Moving to extra-focal position")
+
+        # Hexapod offsets are relative, so need to move 2x the offset
+        # to get from the intra- to the extra-focal position.
+        z_offset = -(self.dz * 2.0)
+        await self.mtcs.move_camera_hexapod(x=0, y=0, z=z_offset, u=0, v=0)
```
tribeiro, line 211: See my comment above about using `offset_camera_hexapod` instead.

python/lsst/ts/standardscripts/maintel/take_triplet_comcam.py
```diff
@@ -0,0 +1,248 @@
...
+            group_id=self.group_id,
+            filter=self.filter,
+            reason="EXTRA" + ("" if self.reason is None else f"_{self.reason}"),
+            program=self.program,
+        )
+
+        self.log.debug("Moving to in-focus position")
+
+        # Move the hexapod back to in focus position
+        await self.mtcs.move_camera_hexapod(x=0, y=0, z=self.dz, u=0, v=0)
```
tribeiro, line 227: See comment above about using `offset_camera_hexapod` instead.

tribeiro (commented):
Looks pretty good. I am ready to approve this PR, however, I would like to make sure you check my comment about using `offset_camera_hexapod` instead of `move_camera_hexapod`. I think using the later like you are doing now considerably reduces the use case for this script. You probably want to use `offset` instead.

tribeiro (approved):
Looks good!

PR #197, opened 2024-07-13
Tickets/dm 44766

elanaku:
For https://rubinobs.atlassian.net/browse/DM-44766
Adds three scripts for the AT dome:

- enable dome following
- disable dome following
- slew dome

### DIFF
--- a/python/lsst/ts/standardscripts/auxtel/atdome/slew_dome.py
+++ b/python/lsst/ts/standardscripts/auxtel/atdome/slew_dome.py
@@ -1,0 +1,91 @@
+# This file is part of ts_standardscripts
+#
+# Developed for the LSST Telescope and Site Systems.
+# This product includes software developed by the LSST Project
+# (https://www.lsst.org).
+# See the COPYRIGHT file at the top-level directory of this distribution
+# for details of code ownership.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program. If not, see <https://www.gnu.org/licenses/>.
+__all__ = ["SlewDome"]
+
+import yaml
+from lsst.ts import salobj
+from lsst.ts.observatory.control.auxtel.atcs import ATCS
+
+
+class SlewDome(salobj.BaseScript):
+    """Run slew dome on ATCS.
+
+    Parameters
+    ----------
+    index : `int`
+        Index of Script SAL component.
+
+    Notes
+    -----
+    **Checkpoints**
+
+    None
+
+    """
+
+    def __init__(self, index):
+        super().__init__(
+            index=index,
+            descr="Slew the ATDome.",
+        )
+
+        self.atcs = None
+
+        self.az = None
+
+    @classmethod
+    def get_schema(cls):
+        yaml_schema = """
+            $schema: http://json-schema/draft-07/schema#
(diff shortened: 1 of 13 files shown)


python/lsst/ts/standardscripts/auxtel/atdome/slew_dome.py
```diff
@@ -0,0 +1,97 @@
...
+        """
+        self.log.debug(f"Configuration: {config}")
+
+        self.az = config.az
+
+        if self.atcs is None:
+            self.atcs = ATCS(domain=self.domain, log=self.log)
+
+    def set_metadata(self, metadata):
+        metadata.duration = 300.0
```
edennihy, line 90: I think we can do a better estimate here. The atdome moves at a speed of 6 deg/s (https://github.com/lsst-ts/ts_atdome/blob/3c248f68cabcf03ae431dae66fc2ca7a47edc826/python/lsst/ts/atdome/mock_controller.py#L89). You could create a short method to calculate the slew distance and use that to provide a better estimate. Otherwise, I think you can consider a maximum 60s (corresponding to 360 deg of movement).

python/lsst/ts/standardscripts/auxtel/atdome/slew_dome.py
edennihy, line 1: I don't think we want the "null" option as a valid configuration for the az in this case. The most likely scenario of a user passing a "null" azimuth would be if they forgot to add the config. And as it stands, the script would still pass, just with a warning. If they didn't see that warning, there's a good chance they will assume the dome has slewed to the position they expected (even though they forgot the config). 

My recommendation is to leave the az as a required configuration, but remove the "null" option and do not set a default value. This way the script will fail at configuration stage if an azimuth value is not passed.

edennihy (changes requested):
This is pretty close, I've added some minor changes requested to the enable/disable scripts, as well as a recommendation to change to the slew_dome script configuration stage.

python/lsst/ts/standardscripts/auxtel/atdome/slew_dome.py
elanaku: I just changed it to 60s. I could write the routine if you prefer.

python/lsst/ts/standardscripts/auxtel/atdome/slew_dome.py
```diff
@@ -64,8 +64,6 @@ def get_schema(cls):
                 description: Azimuth position (in degrees) to slew the dome to.
                 anyOf:
```
edennihy, line 65: Thanks for adopting this recommendation. Since it only allows the one type, no need to do the anyOf here.

python/lsst/ts/standardscripts/auxtel/atdome/slew_dome.py
```diff
@@ -0,0 +1,95 @@
...
+        self.az = None
+
+    @classmethod
+    def get_schema(cls):
+        yaml_schema = """
+            $schema: http://json-schema/draft-07/schema#
+            $id: https://github.com/lsst-ts/ts_externalscripts/auxtel/SlewDome.yaml
+            title: SlewDome v1
+            description: Configuration for SlewDome.
+            type: object
```
edennihy, line 61: You can replace this with `type: number`

python/lsst/ts/standardscripts/auxtel/atdome/slew_dome.py
```diff
@@ -0,0 +1,95 @@
...
+        self.az = config.az
+
+        if self.atcs is None:
+            self.atcs = ATCS(domain=self.domain, log=self.log)
+
+    def set_metadata(self, metadata):
+        metadata.duration = 60.0
+
+    async def run(self):
+        if self.az is None:
```
edennihy, line 91: Since we updated the configuration, this if/else loop is no longer needed. If the user does not pass the `az`, the script will fail at configuration stage.

edennihy (approved):
A few more minor comments to consider following the latest updates. Please also remember to squash your commits before merging. I think one commit statement per script (rather than one per file) is sufficient.

python/lsst/ts/standardscripts/auxtel/atdome/slew_dome.py
edennihy: That's okay, I don't think 60s is egregious.

PR #198, opened 2024-07-15
Tickets/dm 44192: Update version history notes and towncrier to link towards cloud jira instead of local

edennihy:


tribeiro (approved):
looks good

PR #200, opened 2024-07-15
Tickets/dm 45229

dsanmartim:
Hi @tribeiro 

Here is the first version of it. I added a unit test to test the obs_ids (with a mock), but let me know if you think it is unnecessary.

Tks!

### DIFF
--- a/python/lsst/ts/standardscripts/base_block_script.py
+++ b/python/lsst/ts/standardscripts/base_block_script.py
@@ -27,6 +27,7 @@ import hashlib
 import io
 import json
 import os
+import re
 from datetime import datetime
 
 import yaml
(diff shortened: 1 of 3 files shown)


python/lsst/ts/standardscripts/base_block_script.py
```diff
@@ -162,16 +178,23 @@ async def get_obs_id(self) -> str | None:
             return None
 
         try:
-            ticket_id = abs(int(ticket_id_str))
+            ticket_id = abs(int(block_id))
             image_server_url = IMAGE_SERVER_URL[site]
+
+            # Determine if it's a Block test case or Block ticket
+            if match.groupdict()["block_test_case"] is not None:
+                block_type = "BlockTestCase"
```
tribeiro, line 186: This should be `BlockT` only.

tribeiro (approved):
Looks good. But see my inline comment about the `block_type`. The camera index server assumes it is `BlockT` and not `BlockTestCase`.

PR #203, opened 2024-07-19
DM-45351: Add metadata to latiss_checkout script 

hsinfang:


python/lsst/ts/standardscripts/auxtel/daytime_checkout/latiss_checkout.py
```diff
@@ -86,12 +86,15 @@ def get_schema(cls):
 
     async def configure(self, config):
         # This script does not require any configuration
-        pass
+        self.program = "LatissCheckout"
```
edennihy, line 89: Can we use the Test Case name here instead "BLOCK-T17"?

edennihy (commented):
We use the towncrier system for automatically generating version history, please see the documentation here on how to add a feature "news" fragment https://github.com/lsst-ts/ts_standardscripts/tree/develop/doc/news.

edennihy:
Also, please run black to reformat these changes. Once that is ready, if you could squash these into a single commit (then a separate commit for the Add new fragment), I think we'll be all set!

hsinfang:
Thank you for reviewing and for the tips working here! 

The survey name was updated, the commits were squashed, and the new fragment was added in a separate commit.

edennihy (approved):
Great, this looks good! Once CI passes you can merge.

PR #204, opened 2024-07-21
Tickets/dm 45350

elanaku:
https://rubinobs.atlassian.net/browse/DM-45350

### DIFF
--- a/tests/test_maintel_take_stuttered_comcam.py
+++ b/tests/test_maintel_take_stuttered_comcam.py
@@ -1,0 +1,45 @@
+# This file is part of ts_standardscripts
+#
+# Developed for the LSST Telescope and Site Systems.
+# This product includes software developed by the LSST Project
+# (https://www.lsst.org).
+# See the COPYRIGHT file at the top-level directory of this distribution
+# for details of code ownership.
+#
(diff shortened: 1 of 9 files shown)


tests/test_maintel_take_stuttered_comcam.py
tribeiro, line 1: I suggest leaving only the `test_executable` test in this script. Everything else is already covered in the `take_stuttered_latiss`.

tests/test_maintel_take_stuttered_lsstcam.py
tribeiro, line 1: Same comment as above, leave only the `test_executable` test, everything else is tested on `latiss` version.

tribeiro (commented):
Looks good, I just have a few comments inline.

tribeiro:
I forgot to mention but the tests still needs to have the `basic_make_script` method. 

You can do one of the 2 options:

1. Add `base_make_script` back to each test.
2. Add the tests to the `tests/test_auxtel_take_stuttered_latiss.py` class. 


Also, I think you branch needs to be rebased to develop.

tribeiro (approved):
Looks good! thanks

PR #205, opened 2024-07-23
tickets/DM-44629: Add SAL Script to enable ATAOS corrections

cvillalon:
Add SAL script to enable ATAOS corrections but skipping having the corrections as a configuration option.


tribeiro (approved):
Looks good! thanks!

PR #208, opened 2024-07-31
Add DisableATAOSCorrections script for auxtel

MarcoRocchietti:
Add SAL script to disable ATAOS corrections but skipping having the corrections as a configuration option.


tribeiro (approved):
Looks good!

PR #209, opened 2024-08-01
Pass filter name and rotator angle to OFC in `base_closed_loop.py`

gmegh:


python/lsst/ts/standardscripts/maintel/base_close_loop.py
```diff
@@ -442,6 +446,11 @@ async def arun(self, checkpoint: bool = False) -> None:
             # Run the operational mode handler function.
             await self.operation_model_handlers[self.mode]()
 
+            # Compute the rotation angle
```
tribeiro, line 449: you are not computing the rotation angle, you are retrieving it.

python/lsst/ts/standardscripts/maintel/base_close_loop.py
```diff
@@ -452,7 +461,7 @@ async def arun(self, checkpoint: bool = False) -> None:
             )
 
             # Compute ts_ofc offsets
-            dof_offset = await self.compute_ofc_offsets()
+            dof_offset = await self.compute_ofc_offsets(rotation_angle)
```
tribeiro, line 464: you need to pass the value from the topic:

```
dof_offset = await self.compute_ofc_offsets(rotation_angle.actualPosition)
```

python/lsst/ts/standardscripts/maintel/base_close_loop.py
```diff
@@ -384,14 +384,18 @@ async def handle_cwfs_mode(self) -> None:
...
 
-    async def compute_ofc_offsets(self) -> None:
-        """Compute offsets using ts_ofc."""
+    async def compute_ofc_offsets(self, rotation_angle: float) -> None:
+        """Compute offsets using ts_ofc.
 
+        Parameters
+        ----------
+        rotation_angle : `float`
+            Rotation angle of the camera.
```
tribeiro, line 393: Add expected units.

python/lsst/ts/standardscripts/maintel/base_close_loop.py
```diff
@@ -442,6 +446,11 @@ async def arun(self, checkpoint: bool = False) -> None:
             # Run the operational mode handler function.
             await self.operation_model_handlers[self.mode]()
 
+            # Compute the rotation angle
+            rotation_angle = await self.mtcs.rem.mtrotator.tel_rotation.next(
```
tribeiro, line 450: just one question. The data is taken before this point here. Do you care that the rotator will have change positions? you may want to have a different strategy where you get the rotation angle before/after the data is being taken and pass an average value?

tribeiro (commented):
Here some initial comments.

python/lsst/ts/standardscripts/maintel/base_close_loop.py
gmegh: I added the average as you suggest

python/lsst/ts/standardscripts/maintel/base_close_loop.py
```diff
@@ -452,7 +471,7 @@ async def arun(self, checkpoint: bool = False) -> None:
             )
 
             # Compute ts_ofc offsets
-            dof_offset = await self.compute_ofc_offsets()
+            dof_offset = await self.compute_ofc_offsets(rotation_angle.actualPosition)
```
tribeiro, line 474: ok, so now `rotation_angle` is a float so you should just pass `rotation_angle`

tests/test_maintel_close_loop_lsstcam.py
```diff
@@ -55,6 +55,8 @@ async def basic_make_script(self, index):
         # MTCS mocks
         self.script.mtcs.assert_all_enabled = unittest.mock.AsyncMock()
         self.script.mtcs.move_camera_hexapod = unittest.mock.AsyncMock()
+        self.script.mtcs.rem.mtrotator = unittest.mock.AsyncMock()
+        self.script.mtcs.rem.mtrotator.tel_rotation.next = self.return_rotation_angle()
```
tribeiro, line 59: You should do something like:


```
self.script.mtcs.rem.mtrotator.configure_mock(
      **{
            "tel_rotation.next": types.SimpleNamespace(actualPosition=0.0),
         }
)
```

tribeiro (commented):
A couple more comments. I think once you take these into consideration, the CI will pass and I will be ready to approve it.

tribeiro (approved):
Looks good!

PR #210, opened 2024-08-09
tickets/DM-45266: Add configuration option to pass focus sweep steps as array to BaseFocusSweep

cvillalon:


python/lsst/ts/standardscripts/base_focus_sweep.py
```diff
@@ -120,6 +120,12 @@ def get_schema(cls) -> dict:
                 description: Number of steps to take inside the focus window.
                 type: number
                 minimum: 3
+              focus_steps:
```
edennihy, line 123: Maybe we could use a more descriptive name and description here? As-is, it kind of looks like it is something we would use with focus_window and n_steps. Maybe instead of `focus_steps`, this could be `focus_step_sequence`?

python/lsst/ts/standardscripts/base_focus_sweep.py
```diff
@@ -120,6 +120,12 @@ def get_schema(cls) -> dict:
...
+                description: Steps to take for the focus sweep.
```
edennihy, line 124: We can then update this description to say "User-provided sequence of focus steps to take for the focus sweep, used for unevenly spaced steps."

python/lsst/ts/standardscripts/base_focus_sweep.py
```diff
@@ -120,6 +120,12 @@ def get_schema(cls) -> dict:
...
+                type: array
+                items:
+                  type: number
+                minItems: 3
```
edennihy, line 128: I think technically minItems could be 2.

python/lsst/ts/standardscripts/base_focus_sweep.py
```diff
@@ -142,7 +148,14 @@ def get_schema(cls) -> dict:
...
                   type: string
-            required: ["axis", "focus_window", "n_steps"]
+            oneOf:
+                - required:
+                    - axis
+                    - focus_window
+                    - n_steps
+                - required:
+                    - axis
+                    - focus_steps
```
edennihy, line 158: update parameter name to match above.

python/lsst/ts/standardscripts/base_focus_sweep.py
```diff
@@ -183,6 +196,15 @@ async def configure(self, config: types.SimpleNamespace) -> None:
                         f"{self.camera.components_attr}. Ignoring."
                     )
 
+        if hasattr(config, "focus_steps"):
+            config.focus_window = config.focus_steps[-1] - config.focus_steps[0]
+            config.n_steps = len(config.focus_steps)
```
edennihy, line 201: If you update the name above, we will have to assign `config.focus_steps` here.

tests/test_auxtel_focus_sweep_latiss.py
```diff
@@ -81,6 +81,51 @@ async def test_configure(self):
             assert self.script.config.grating == 1
             assert self.script.config.n_images_per_step == 1
 
+    async def test_configure_focus_steps(self):
+        config = {
+            "axis": "x",
+            "focus_steps": [0, 100, 200, 300, 400],
```
edennihy, line 87: This is kind of a picky comment, but the scale here is pretty far off from what we normally do with the real hexapods. Focus steps are provided to the axes in units of mm, so we usually do steps of 0.1mm or less. Would you mind updating these tests to something like [0, 0.1, 0.2, 0.3, 0.4]?

edennihy (changes requested):
Hi Carolina, thanks for the quick turnaround and for adding the unit tests. I left a few comments for your consideration.

python/lsst/ts/standardscripts/base_focus_sweep.py
cvillalon: Ok. I copied this value from ``n_steps``. Should I change ``minItems`` for ``n_steps`` too?

python/lsst/ts/standardscripts/base_focus_sweep.py
cvillalon: I've decided to use ``focus_step_sequence`` all along so I changed every occurrence of ``focus_steps``. I hope that's ok.

tests/test_auxtel_focus_sweep_latiss.py
cvillalon: Great comment! I was unsure of what values to use here so I looked at the rest of the tests that were already written and tried to use similar ones. I've changed all tests written by me to use these values and changed simple ``assert``s to ``assertAlmostEqual`` to deal with floating point representation. Let me know if you want me to update older tests too.

python/lsst/ts/standardscripts/base_focus_sweep.py
edennihy: Sure, I think that makes sense.

tests/test_auxtel_focus_sweep_latiss.py
edennihy: Actually I checked with Tiago on this one and the simonyi hexapod offsets are given in microns, not mm like they are in AuxTel. So hundreds of microns actually makes sense for some of these unit tests (the comcam and lsstcam versions). Sorry, I was only thinking of auxtel when I made this comment.

edennihy (approved):

PR #211, opened 2024-08-10
DM-45637: Implement block related functionality in the ScriptQueue and Scripts.

tribeiro:
This PR is mainly to add backward compatibility for the unit tests. 

Once we roll this feature to the summit we can start removing the BaseBlockScript functionality.


wvreeven (approved):
LGTM

PR #212, opened 2024-08-13
Tickets/dm 45747: In base_track_target script, remove azimuth limits from configuration schema

edennihy:


tribeiro (approved):
Looks good

PR #213, opened 2024-08-14
tickets/DM-45774: Update config in base_focus_sweep to avoid re-centering user provided steps

cvillalon:
Update configuration and tests.


edennihy (approved):
Looks good! Thanks for the quick turnaround.

PR #215, opened 2024-08-14
New SalScript for Power On for TunableLaser

parfa30:
This will be used for functional testing and commissioning of TunableLaser before the rest of the calibration system is instealled.

### DIFF
--- a/python/lsst/ts/standardscripts/maintel/calibration/power_on_tunablelaser.py
+++ b/python/lsst/ts/standardscripts/maintel/calibration/power_on_tunablelaser.py
@@ -1,0 +1,154 @@
+# This file is part of ts_standardscripts
+#
+# Developed for the LSST Telescope and Site Systems.
+# This product includes software developed by the LSST Project
+# (https://www.lsst.org).
+# See the COPYRIGHT file at the top-level directory of this distribution
+# for details of code ownership.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+__all__ = ["PowerOnTunableLaser"]
+
+
+import yaml
+from lsst.ts import salobj
+from lsst.ts.observatory.control.maintel.mtcalsys import MTCalsys
+
+
+class PowerOnTunableLaser(salobj.BaseScript):
+    """Starts propagating the Tunable Laser for functional
+    testing.
+
+    Parameters
+    ----------
+    index : `int`
+        Index of Script SAL component.
+
+    """
+
+    def __init__(self, index):
+        super().__init__(
+            index=index,
+            descr="Power On Tunable Laser",
+        )
+
+        self.laser = None
+        self.mtcalsys = None
+
+    @classmethod
+    def get_schema(cls):
+        schema_yaml = """
+            $schema: http://json-schema.org/draft-07/schema#
+            $id: https://github.com/lsst-ts/ts_standardscripts/maintel/calibrations/power_on_tunablelaser.yaml
+            title: PowerOnTunableLaser v1
+            description: Configuration for PowerOnTunableLaser.
+              Each attribute can be specified as a scalar or array.
+              All arrays must have the same length (one item per image).
+            type: object
+            properties:
+              sequence_name:
+                description: Name of sequence in MTCalsys
+                type: string
+                default: laser_functional
+
+            additionalProperties: false
+        """
+        return yaml.safe_load(schema_yaml)
+
+    def set_metadata(self, metadata):
+        metadata.duration = 30
+
+    async def configure(self, config):
+        """Configure the script.
+
+        Parameters
+        ----------
+        config : ``self.cmd_configure.DataType``
+
+        """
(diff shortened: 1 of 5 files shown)


parfa30:
Won't pass jenkins until this is merged: https://github.com/lsst-ts/ts_observatory_control/pull/160

python/lsst/ts/standardscripts/maintel/calibration/power_on_tunablelaser.py
```diff
@@ -0,0 +1,178 @@
...
+
+        await self.checkpoint("Starting laser propagation")
+        await self.start_propagation_on()
+
+    async def start_propagation_on(self):
+        """Starts propagation of the laser"""
+
+        await self.laser.cmd_startPropagateLaser.start(timeout=self.laser_warmup)
+
+    async def configure_tunablelaser(self):
```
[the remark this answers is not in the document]
parfa30: Here is the issue with moving all of this to MTCalsys and why I want to define it in a script. There will be scripts in which the wavelength and optical configuration are cycled through in loops. It's not a static configuration and they will be slightly different in each script.

python/lsst/ts/standardscripts/maintel/calibration/power_on_tunablelaser.py
```diff
@@ -0,0 +1,178 @@
...
+    testing.
+
+    Parameters
+    ----------
+    index : `int`
+        Index of Script SAL component.
+
+    """
+
+    def __init__(self, index, add_remotes: bool = True):
```
tribeiro, line 42: Please, drop the `add_remotes` here... We no longer use this, and it doesn't seems like you are using it.

python/lsst/ts/standardscripts/maintel/calibration/power_on_tunablelaser.py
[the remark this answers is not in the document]
tribeiro: I agree with @edennihy this should go on `MTCalsys`, especially, as you mention "There will be scripts"... so, if you are using this routine in more than one place it must absolutely be in the control class.

You can always make these arguments to methods in `MTCalsys`, then here you would call it like:

```
await self.mtcalsys.configure_tunable_laser(wavelength=self.wavelength, optical_configuration=optical_configuration)
```

And in the next script you write that needs to do this on a loop you would do:

```
for wavelength, optical_configuration in self.get_tunable_laser_configurations():
    await self.mtcalsys.configure_tunable_laser(wavelength=self.wavelength, optical_configuration=optical_configuration)
```

and won't have to copy paste what you did here.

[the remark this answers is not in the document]
parfa30: Ok. I have opened this ticket and will make the necessary changes to observatory_control and then will finish this script: https://rubinobs.atlassian.net/browse/DM-46299. I'll ask you both for review of that ticket once I can get to it.

python/lsst/ts/standardscripts/maintel/calibration/power_on_tunablelaser.py
```diff
@@ -0,0 +1,178 @@
...
+        )
+        await self.laser.cmd_setOpticalConfiguration.set_start(
+            configuration=self.optical_configuration, timeout=self.long_timeout
+        )
+        if self.mode == TunableLaser.LaserDetailedState.NONPROPAGATING_CONTINUOUS_MODE:
+            await self.laser.cmd_setContinuousMode.start(timeout=self.long_timeout)
+        elif self.mode == TunableLaser.LaserDetailedState.NONPROPAGATING_BURST_MODE:
+            await self.laser.cmd_setBurstMode.start(timeout=self.long_timeout)
+
+        params = await self.get_laser_parameters()
```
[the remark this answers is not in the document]
parfa30: I wonder how that passed :/

python/lsst/ts/standardscripts/maintel/calibration/power_on_tunablelaser.py
```diff
@@ -0,0 +1,166 @@
...
+        await self.checkpoint("Configuring TunableLaser")
+        await self.configure_tunablelaser()
+
+        await self.checkpoint("Starting laser propagation")
+        await self.start_propagation_on()
+
+    async def start_propagation_on(self):
+        """Starts propagation of the laser"""
+
+        await MTCalsys.laser_start_propagate.start()
```
edennihy, line 120: this should be `self.mtcalsys.laser_start_propagate()`

python/lsst/ts/standardscripts/maintel/calibration/power_on_tunablelaser.py
```diff
@@ -0,0 +1,166 @@
...
+
+    async def configure_tunablelaser(self):
```
edennihy, line 122: Since you defined this method in the MTCalSys class, I don't think you need the configure_tunablelaser(self) method anymore. You can just move these lines directly into the run method. They would then be:

`self.mtcalsys.setup_laser()`

python/lsst/ts/standardscripts/maintel/calibration/power_on_tunablelaser.py
```diff
@@ -0,0 +1,166 @@
...
+
+        self.log.info(
+            f"Laser Configuration is {params[0]}, \n"
+            f"wavelength is {params[1]}, \n"
+            f"Interlock is {params[2]}, \n"
+            f"Burst mode is {params[3]}, \n"
+            f"Cont. mode is {params[4]}"
+        )
+
+    async def get_tunablelaser_parameters(self):
```
edennihy, line 141: While we're adding methods, this seems like a super useful method to add to the MTCalSys class since I suspect it will get called in most tunable laser scripts.

parfa30: Good idea

edennihy (approved):
Approved, thanks for all the changes! Please squash your commits (you can squash these all down to one) when you rebase the branch.

PR #217, opened 2024-08-23
tickets/DM-42061: Move the AT to CO2 cleanup position

cvillalon:
Add SAL script to move auxiliary telescope to O2 cleanup position.

### DIFF
--- a/python/lsst/ts/standardscripts/auxtel/prepare_for/co2_cleanup.py
+++ b/python/lsst/ts/standardscripts/auxtel/prepare_for/co2_cleanup.py
@@ -1,0 +1,146 @@
+# This file is part of ts_standardscripts
+#
+# Developed for the LSST Telescope and Site Systems.
+# This product includes software developed by the LSST Project
+# (https://www.lsst.org).
+# See the COPYRIGHT file at the top-level directory of this distribution
+# for details of code ownership.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+__all__ = ["PrepareForCO2Cleanup"]
+
+import yaml
+from lsst.ts import salobj
+from lsst.ts.observatory.control.auxtel import ATCS, ATCSUsages
+
+
+class PrepareForCO2Cleanup(salobj.BaseScript):
+    """Put AT in CO2 cleanup position.
+
+    This script will slew the auxiliary telescope to the CO2 cleanup position.
+    Telescope will be left in CO2 cleanup position with tracking disabled and
+    mirror cover open.
+
+    Parameters
+    ----------
+    index : `int`
+        Index of Script SAL component.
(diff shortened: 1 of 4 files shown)


python/lsst/ts/standardscripts/auxtel/slew_to_O2_cleanup_position.py
```diff
@@ -0,0 +1,142 @@
...
+                    description: Timeout for slew procedure (in seconds).
+                    type: number
+                    default: 180.0
+                ignore:
+                    description: ATCS components to ignore in availability check.
+                    type: array
+                    items:
+                        type: string
+            required:
+                - az
```
tribeiro, line 90: Instead of making these required, let's add defaults that are sensible. I think for now let's start with az = 0 and el = 20 and I will verify if those are ok later.

tribeiro (commented):
Hi @cvillalon , I should have told you this before but let's call this script `o2_cleanup.py` and put it inside the `auxtel/prepare_for` directory.. 

Sorry about that!

Otherwise it all looks good!

tribeiro (commented):
Also, this should be "co2" cleaning not o2 cleaning.. hehe..

cvillalon:
Thanks @tribeiro , I've pushed the changes addressing your comments

python/lsst/ts/standardscripts/auxtel/prepare_for/co2_cleanup.py
```diff
@@ -0,0 +1,141 @@
...
+
+    async def run(self):
+        await self.atcs.assert_all_enabled()
+
+        self.log.debug(
+            "Slew telescope to CO2 cleanup position. "
+            f"Az: {self.config.az}, "
+            f"El: {self.config.el}, "
+            f"Rot: {self.config.rot_tel}"
+        )
```
tribeiro, line 133: A couple things I forgot here, before moving the telescope, we need to enable ATAOS corrections. This will put pressure behind the mirror and make sure it is not moving on the states supports. you can enable it using

```
await atcs.enable_ataos_corrections()
```

python/lsst/ts/standardscripts/auxtel/prepare_for/co2_cleanup.py
```diff
@@ -0,0 +1,141 @@
...
+        await self.atcs.point_azel(
+            az=self.config.az,
+            el=self.config.el,
+            rot_tel=self.config.rot_tel,
+            target_name=self.config.target_name,
+            wait_dome=self.config.wait_dome,
+            slew_timeout=self.config.slew_timeout,
+        )
```
tribeiro, line 141: After moving the telescope you should disable the ATAOS corrections and open the mirror covers for cleaning, e.g.

```
await atcs.disable_ataos_corrections()
await atcs.open_m1_cover()
```

tribeiro (approved):
This looks good. I do have a couple more additions to the operations. I am going to approve the PR but please, make sure you include the additional operations before merging.

python/lsst/ts/standardscripts/auxtel/prepare_for/co2_cleanup.py
cvillalon: I see that ``atcs.open_m1_cover`` requires a minimum elevation of 70 degrees, so I think it would be best to change the default elevation in the configuration to this value. Is that ok?

tribeiro: Right! We should probably open m1 before moving then...

cvillalon: Is it ok to move the telescope with the mirror cover open?

tribeiro: Yes

PR #221, opened 2024-09-20
Tickets/dm 45817

dsanmartim:
Hi @tribeiro 

I implemented this in a way that maintains backward compatibility, ensuring the script continues to work in the same way in existing json blocks.

### DIFF
--- a/doc/news/DM-45817.feature.rst
+++ b/doc/news/DM-45817.feature.rst
@@ -1,0 +1,6 @@
+In  ``offset_camera_hexapod.py`` and ``offset_m2_hexapod.py``:
+
+- Added an option to reset hexapod offsets before applying new ones.
+  This ensures that the offsets are first reset to zero before applying the user-provided offsets.
+- Added an option to reset hexapod offsets to zero without performing any additional actions.
+  This allows the user to reset the hexapod offsets to zero without applying new offsets.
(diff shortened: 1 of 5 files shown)


python/lsst/ts/standardscripts/maintel/offset_camera_hexapod.py
```diff
@@ -78,6 +78,18 @@ def get_schema(cls):
               v:
                 type: number
                 description: Ry offset (deg).
+              reset_axis:
```
tribeiro, line 81: I think the naming here is backwards, `reset_only` should be the boolean and `reset_axis` the name of the axis you want to reset.

dsanmartim: I made it in this way to give flexibility, so the user can reset the axis first before applying the offsets (maybe this property should be called `reset_first`). For the `reset_only` property, I focused on giving flexibility to reset any axis the user may want, instead of resetting all of them (which would be case if this is a boolean).

dsanmartim: I refactored the code to have only the `reset_axis` property to deal with all possible scenarios to reset the axes. Thanks for your suggestions!

tribeiro (approved):
Ok, I only have some minor typos comments in the news fragments. Everything else looks ok.

doc/news/DM-45817.feature.rst
```diff
@@ -0,0 +1,3 @@
+In  ``offset_camera_hexapod.py``and ``offset_m2_hexapod.py``:
+- add an option to reset hexapod offsets to zero. In this case, it only reset offset and don't do anything else.
+- add an option to reset hexapod offsets before applying new offsets. In this case, it reset offset before applying new offsets.
```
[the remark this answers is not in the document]
dsanmartim: Tks. The messages were really bad actually. I just improved them, taking your suggestions into account.

PR #227, opened 2024-10-08
Add note configuration parameter to `close_loop_comcam.py` script

gmegh:


python/lsst/ts/standardscripts/maintel/base_close_loop.py
```diff
@@ -175,6 +175,12 @@ def get_schema(cls) -> typing.Dict[str, typing.Any]:
                   - type: string
                   - type: "null"
                 default: null
+              note:
+                description: Optional note for taking the data.
```
edennihy, line 179: Please update the description to be consistent with how `note` is defined in other data taking scripts e.g. https://github.com/lsst-ts/ts_standardscripts/blob/995ba83673fe10e47c253582b19bb63c3f72b3bd/python/lsst/ts/standardscripts/maintel/take_triplet_comcam.py#L101

edennihy (approved):
Please consider this comment before merging.

PR #229, opened 2024-10-13
Tickets/dm 46179 Extend TCS readiness check to other image types beyond OBJECT

iglesu:
https://rubinobs.atlassian.net/browse/DM-46179

Before I keep going I want to make sure I'm not deviating from the intended development objectives. Specially for the unit tests. Thanks!

### DIFF
--- a/python/lsst/ts/standardscripts/maintel/take_image_comcam.py
+++ b/python/lsst/ts/standardscripts/maintel/take_image_comcam.py
@@ -22,7 +22,9 @@
 __all__ = ["TakeImageComCam"]
 
 import yaml
+from lsst.ts.observatory.control.maintel import MTCS
 from lsst.ts.observatory.control.maintel.comcam import ComCam, ComCamUsages
+from lsst.ts.observatory.control.maintel.mtcs import MTCSUsages
 
 from ..base_take_image import BaseTakeImage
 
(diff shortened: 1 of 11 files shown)


python/lsst/ts/standardscripts/maintel/take_image_comcam.py
```diff
@@ -48,8 +49,13 @@ def __init__(self, index):
 
         self.config = None
 
+        self.mtcs = MTCS(self.domain, log=self.log)
```
tribeiro, line 52: Same comment as above for the auxtel script, but it is a lot more critical for MT, you need to limit the usages or it will use way too many resources. It should be something like:

```
self.mtcs = MTCS(self.domain, log=self.log, intended_usage=MTCSUsages.Slew)
```

python/lsst/ts/standardscripts/maintel/take_image_lsstcam.py
```diff
@@ -48,8 +49,13 @@ def __init__(self, index):
 
         self.config = None
 
+        self.mtcs = MTCS(self.domain, log=self.log)
```
tribeiro, line 52: Please restrict the resources.

tribeiro (approved):
Looks good..

PR #230, opened 2024-10-15
Tickets/DM-46864: Take AOS Sequence

gmegh:


tribeiro (commented):
Consider creating an enumeration to hold the valid values for the `mode`. 

Then you can compare enum values instead of strings, which will make it more readable and easy to update.

you could have something like:

```
class Mode(enum.IntEnum):
    TRIPLET = enum.auto()
    INTRA = enum.auto()
    EXTRA = enum.auto()
```

Then in the configuration you could do something like this.

```
schema_yaml = f"""
...
              mode:
                description: >-
                    Mode of operation. Options are 'triplet' (default), 'intra' or 'extra'.
                type: str
                default: 'triplet'
                enum: {[mode for mode in Mode]}
```

Then, you will have to covert the value to a mode enum:

```
self.mode = getattr(Mode, config.mode)
```

and then compare mode to the enum values instead of strings, e.g:

```
if self.mode == Mode.TRIPLET or self.mode == Mode.INTRA:
    ...
```
```

File: python/lsst/ts/standardscripts/maintel/take_aos_sequence_comcam.py [hunk 1]
```diff
@@ -169,12 +187,13 @@ def set_metadata(self, metadata: salobj.type_hints.BaseMsgType) -> None:
             directly.
         """
         # Estimated duration is maximum number of iterations multiplied by
-        # 3 (data triplet) multiplied by the time it takes to take an image
+        # 3 or 2 multiplied by the time it takes to take an image
         # plus estimation on reading out the images (10s)
+        number_of_images = 3 if self.mode == "triplet" else 2
```
tribeiro, line 192: Should be 

```
number_of_images = 3 if self.mode == Mode.TRIPLET else 2
```

python/lsst/ts/standardscripts/maintel/take_aos_sequence_comcam.py
```diff
@@ -224,88 +243,97 @@ async def configure_tcs(self) -> None:
...
-    ) -> None:
-        """Take triplet (intra focal, extra focal and in-focus)
-        image sequence.
-        """
-        self.log.debug("Moving to intra-focal position")
-
-        await self.mtcs.offset_camera_hexapod(x=0, y=0, z=self.dz, u=0, v=0)
-
-        self.log.info("Taking intra-focal image")
+    async def take_out_of_focus_sequence(self) -> None:
```
tribeiro, line 246: Maybe `take_aos_sequence`?

python/lsst/ts/standardscripts/maintel/take_aos_sequence_comcam.py
```diff
@@ -224,88 +243,97 @@ async def configure_tcs(self) -> None:
...
+            Target z position to move to (relative offset).
+        """
+        z_move = target_z - self.current_z_position  # Calculate relative move
+        await self.mtcs.offset_camera_hexapod(x=0, y=0, z=z_move, u=0, v=0)
+        self.current_z_position = target_z  # Update current position
+        self.log.debug(f"Moved hexapod to z={target_z} (relative move: {z_move}).")
 
-        # Move the hexapod back to in focus position
-        await self.mtcs.offset_camera_hexapod(x=0, y=0, z=self.dz, u=0, v=0)
+    async def take_cwfs_image(self, reason_suffix: str) -> int:
```
tribeiro, line 308: I don't see the value of having this method here. Why not just call `self.camera.take_cwfs` directly? note that, because this is a method, you need to add `self.supplemented_group_id` as a class attribute. I suggest you get rid of this method (and also remove `self.supplemented_group_id` in favor of a local variable. If you really want to have a method to encapsulate this operation (it does make the inline call leaner), pass `supplemented_group_id` alongside `reason_suffix` instead of relying on a class attribute.

python/lsst/ts/standardscripts/maintel/take_aos_sequence_comcam.py
```diff
@@ -224,88 +243,97 @@ async def configure_tcs(self) -> None:
...
-
     async def run_block(self) -> None:
         """Execute script operations."""
         await self.assert_feasibility()
 
-        for i in range(self.n_triplets):
-            self.log.info(f"Starting triplet {i+1} of {self.n_triplets}")
-            await self.checkpoint(f"triplet {i+1} of {self.n_triplets}")
+        for i in range(self.n_sequences):
+            self.log.info(f"Starting out-of-focus sequence {i+1} of {self.n_sequences}")
```
tribeiro, line 336: Please, replace "out-of-focus" with "aos".

tribeiro (approved):

PR #233, opened 2024-10-17
Tickets/DM-45823: Establish um as the unit for hexapod offsets in BaseFocusSweep and implement conversion to mm for auxtel in FocusSweepLatiss

MarcoRocchietti:


python/lsst/ts/standardscripts/auxtel/focus_sweep_latiss.py
```diff
@@ -102,6 +102,31 @@ def get_schema(cls) -> dict:
...
+    async def configure(self, config):
+        """Configure script.
+
+        Parameters
+        ----------
+        config : `types.SimpleNamespace`
+            Script configuration, as defined by `schema`.
+        """
+        await super().configure(config=config)
+        self.log.debug("Applying unit translation from um to mm for ATHexapod use.")
```
tribeiro, line 114: Please, merge these 3 debug calls together... In our system, calls to `log` will be sent to the control system and stored in the EFD. Having 3 consecutive calls to log is way more expensive than having only one. Also, note that you can use multiline string to format the text, and make it an f-string, e.g.:

```
self.log.debug(f"""Applying unit conversion from um to mm for ATHexapod use.

Original values in um from base class configuration are:

    focus_window = {self.config.focus_window} um
    focus_step_sequence = {self.config.focus_step_sequence} um

Converted values:
    focus_window = {self.config.focus_window} mm
    focus_step_sequence = {self.config.focus_step_sequence} mm
"""
)
```

Note that I replaced "translation" with "conversion" above.

tribeiro (approved):
Looks good, I only have one comment about how you log things in the `configure` file. Please, take a look before merging.

PR #236, opened 2024-10-21
Tickets/dm 46636 - Support Summit Observing Weeks 41-42 of 2024 

b1quint:


b1quint:
The `apply_dof.py` file had conflicts. I reviewed them with @gmegh and we should be fine now.

tribeiro (approved):

PR #237, opened 2024-10-23
Add OCPS option to `base_close_loop.py` script.

gmegh:


python/lsst/ts/standardscripts/maintel/base_close_loop.py
```diff
@@ -202,6 +202,11 @@ def get_schema(cls) -> typing.Dict[str, typing.Any]:
                     Apply OFC corrections after each iteration.
                 type: boolean
                 default: true
+              ocps:
```
tribeiro, line 205: Please, rename this to `use_ocps`.

tribeiro (changes requested):

PR #243, opened 2024-11-14
tickets/DM-47223: Improve `maintel/m1m3/check_hardpoint.py` to run tests concurrently.

cvillalon:


python/lsst/ts/standardscripts/maintel/m1m3/check_hardpoint.py
```diff
@@ -135,10 +140,11 @@ async def run_block(self):
...
 
-        for idx, hp in enumerate(self.hardpoints):
-            await self.checkpoint(f"Testing hardpoint {hp}.")
-            await self.mtcs.run_m1m3_hard_point_test(hp=hp)
-            self.log.info(f"Tests complete: {idx + 1}/{len(self.hardpoints)}")
+        tasks = [
+            self.run_individual_test(idx, hp)
+            for (idx, hp) in enumerate(self.hardpoints)
+        ]
+        await asyncio.gather(*tasks)
```
tribeiro, line 147: Instead of using `asyncio.gather` (which I think will cancel all tasks if one fail), consider using something like `asyncio.as_completed`.

it would be something like this:

```
hp_failures = []
for task in asyncio.as_completed(tasks):
    try:
        await task
    except Exception as e:
        self.log.exception("Hardpoint test failed. Waiting for remaining tests to complete.")
        hp_failures.append(e)

if hp_failure:
    error_message = f"{len(hp_failure)} out of len(tasks) hard point tests failed:\n"
    for error in hp_failure:
        error_message += f"{error:!r}\n"
    raise RuntimeError(error_message)
```

tribeiro (commented):
Check my comment about replacing `asyncio.gather` with `asyncio.as_completed`.

python/lsst/ts/standardscripts/maintel/m1m3/check_hardpoint.py
cvillalon: I'm not so sure that's the case. This is an extract from the [gather documentation](https://docs.python.org/3/library/asyncio-task.html#asyncio.gather):
> If any Task or Future from the _aws_ sequence is _cancelled_, it is treated as if it raised [`CancelledError`](https://docs.python.org/3/library/asyncio-exceptions.html#asyncio.CancelledError) – the `gather()` call is **not** cancelled in this case. This is to prevent the cancellation of one submitted Task/Future to cause other Tasks/Futures to be cancelled.

Nevertheless, I think I should add the `return_exceptions` so I can report them accordingly.

tribeiro (approved):
looks good.

PR #250, opened 2024-12-02
Tickets/DM-47618: Add mechanism to run the m1m3 bump test on the last failed actuators

MarcoRocchietti:
Add `last_failed` option to the `CheckActuators` script to run bump tests on actuators that failed their last test.


tribeiro (approved):
Looks great! thanks!

PR #251, opened 2024-12-04
Tickets/DM-47601: Add sync dof script

gmegh:


python/lsst/ts/standardscripts/maintel/set_dof.py
```diff
@@ -42,11 +47,131 @@ class SetDOF(ApplyDOF):
...
+            Image time.
+        """
+
+        query = f"""
+            SELECT  "imageDate", "imageNumber"
+            FROM "efd"."autogen"."lsst.sal.CCCamera.logevent_endReadout"
+            WHERE imageDate =~ /{day}/ AND imageNumber = {seq}
+        """
+
+        time = await self.client.influx_client.query(query)
```
tribeiro, line 120: I think you need to make some validation here that it returns something... what if the users passed a bad image number?

python/lsst/ts/standardscripts/maintel/set_dof.py
```diff
@@ -42,11 +47,131 @@ class SetDOF(ApplyDOF):
...
+            State of the system.
+        """
+
+        topics = [f"aggregatedDoF{i}" for i in range(50)]
+
+        current_end = time
+        lookback_interval = TimeDelta(1, format="jd")
+        max_lookback_days = 7
+
+        while True:
```
tribeiro, line 143: I am not sure this is a good idea. I would actually be a bit more aggressive here in limiting the look back and search only between the image the user asked and the previous image (something like seq and seq-1), and if there are no event in that time frame you fail the script.

python/lsst/ts/standardscripts/maintel/set_dof.py
```diff
@@ -42,11 +47,131 @@ class SetDOF(ApplyDOF):
...
+                return np.zeros(50)
+        return state.values.squeeze()
+
     async def run(self) -> None:
         """Run script."""
         # Assert feasibility
         await self.assert_feasibility()
 
+        if self.day is not None and self.seq is not None:
+            self.client = EfdClient("summit_efd")
```
tribeiro, line 171: note that you are hard-coding which EFD to use here, which means we won't be able to test this script at BTS or TTS. 

Consider doing something like the `BaseBlockScript` script does. See [here](https://github.com/lsst-ts/ts_standardscripts/blob/cf97e29d73c40665d1aeecfc806a5038bfcec12e/python/lsst/ts/standardscripts/base_block_script.py#L38-L42) and [here](https://github.com/lsst-ts/ts_standardscripts/blob/cf97e29d73c40665d1aeecfc806a5038bfcec12e/python/lsst/ts/standardscripts/base_block_script.py#L162-L177).

python/lsst/ts/standardscripts/maintel/set_dof.py
```diff
@@ -42,11 +55,156 @@ class SetDOF(ApplyDOF):
...
+        Raises
+        ------
+        RuntimeError
+            Unable to connect to EFD.
+        """
         # Assert feasibility
         await self.assert_feasibility()
 
+        if self.day is not None and self.seq is not None:
+            efd_name = await self.get_efd_name()
```
tribeiro, line 203: Why don't you move all this to a method and here you just do:

```
if self.day is not None and self.seq is not None:
    self.dofs = await self.get_last_issued_state()
```

?

tribeiro (approved):

PR #253, opened 2024-12-07
tickets/DM-48018: Script for balanced AOS triplets.

jfcrenshaw:
This enables taking AOS triplets while splitting dz offsets equally between the camera and M2 hexapods. This allows us to generate the equivalent defocus without simultaneously generating spherical aberration.

The new files subclass the original `take_aos_sequence` equivalents with the following small changes:
- New script is called `take_aos_sequence_balanced_comcam.py`
- New class is named `TakeAOSSequenceBalancedComCam`
- Code to set offsets is now:

```python
await self.mtcs.offset_camera_hexapod(x=0, y=0, z=z_offset / 2, u=0, v=0)
await self.mtcs.offset_m2_hexapod(x=0, y=0, z=z_offset / 2, u=0, v=0)
```

Small changes were made to the original to reduce code-duplication for the subclasses.


jfcrenshaw:
Sorry for not being better about linting and formatting before commits. I will squash these if this gets merged.

PR #254, opened 2024-12-10
Tickets/dm 48022

iglesu:
https://rubinobs.atlassian.net/browse/DM-48022 

Deprecate ignore_m1m3 added disable_m1m3_force_balance property
    
    - Deprecate `ignore_m1m3` property.
    - Added new property `disable_m1m3_force_balance` with default `false`.
        Maintains the ability to disable the M1M3 balance system, in case
        the coupling effect between the elevation axis and m1m3
        support system, repeats again, driving the system to a huge
        oscillation
    - Added unit tests for the property deprecation and the new one

### DIFF
--- a/doc/news/DM-48022.feature.rst
+++ b/doc/news/DM-48022.feature.rst
@@ -1,0 +1,5 @@
+Added new property `disable_m1m3_force_balance` with default `false`.
+Maintains the ability to disable the M1M3 balance system, in case
+the coupling effect between the elevation axis and m1m3
+support system, repeats again, driving the system to a huge
+oscillation
--- a/python/lsst/ts/standardscripts/maintel/home_both_axes.py
+++ b/python/lsst/ts/standardscripts/maintel/home_both_axes.py
@@ -21,7 +21,6 @@
 
 __all__ = ["HomeBothAxes"]
 
-import asyncio
 import time
 
 import yaml
@@ -56,10 +55,8 @@ class HomeBothAxes(salobj.BaseScript):
     def __init__(self, index, add_remotes: bool = True):
         super().__init__(index=index, descr="Home both TMA axis.")
 
+        self.disable_m1m3_force_balance = False
         self.home_both_axes_timeout = 300.0  # timeout to home both MTMount axes.
-
-        self.ignore_m1m3 = False
-        self.warn_wait = 10.0
         self.mtcs = None
 
     @classmethod
@@ -72,7 +69,10 @@ class HomeBothAxes(salobj.BaseScript):
             type: object
             properties:
                 ignore_m1m3:
-                    description: Ignore the m1m3 component?
+                    description: Ignore the m1m3 component? (Deprecated property)
+                    type: boolean
+                disable_m1m3_force_balance:
+                    description: Disable m1m3 force balance?
                     type: boolean
                     default: false
             additionalProperties: false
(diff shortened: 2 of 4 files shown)


python/lsst/ts/standardscripts/maintel/home_both_axes.py
```diff
@@ -75,6 +75,10 @@ def get_schema(cls):
                     description: Ignore the m1m3 component?
                     type: boolean
                     default: false
```
tribeiro, line 77: You might want to remove the default value here or it will always trigger the deprecation warning. Also, please add a note in the description that this is going to be deprecated.

python/lsst/ts/standardscripts/maintel/home_both_axes.py
```diff
@@ -107,8 +115,3 @@ async def run(self):
         elapsed_time = end_time - start_time
 
         self.log.info(f"Homing both axes took {elapsed_time:.2f} seconds")
-
-        if not self.ignore_m1m3:
```
tribeiro, line 111: if you disabled it, you need to enable it back here.

python/lsst/ts/standardscripts/maintel/home_both_axes.py
```diff
@@ -56,9 +56,9 @@ class HomeBothAxes(salobj.BaseScript):
     def __init__(self, index, add_remotes: bool = True):
         super().__init__(index=index, descr="Home both TMA axis.")
 
+        self.disable_m1m3_force_balance = False
         self.home_both_axes_timeout = 300.0  # timeout to home both MTMount axes.
 
-        self.ignore_m1m3 = False
         self.warn_wait = 10.0
```
tribeiro, line 62: I think you can remove this attribute now.

python/lsst/ts/standardscripts/maintel/home_both_axes.py
```diff
@@ -83,21 +87,25 @@ async def configure(self, config):
         if self.mtcs is None:
             self.mtcs = MTCS(domain=self.domain, log=self.log)
             await self.mtcs.start_task
-        self.ignore_m1m3 = config.ignore_m1m3
+
+        if hasattr(config, "ignore_m1m3"):
+            warnings.warn(
```
tribeiro, line 92: this warning will be printed to stdout (or stderr, I am not sure) but won't be logged. Please, add a `self.log.warning` call here as well.

doc/news/DM-48022.feature.rst
```diff
@@ -0,0 +1,6 @@
+Deprecate `ignore_m1m3` property.
```
tribeiro, line 1: There's a "removal" type you can probably use to document this one. 

Please, split both of these notes in 2 different files.

doc/news/DM-48022.feature.rst
```diff
@@ -0,0 +1,6 @@
...
+Added new property `disable_m1m3_force_balance` with default `false`.
+    Maintains the ability to disable the M1M3 balance system, in case
```
tribeiro, line 3: I think you can remove this space here before each line. Also, we use "semantic line breaking" so please, make sure each phrase is in one line.

python/lsst/ts/standardscripts/maintel/home_both_axes.py
iglesu: It is removed already, see the "-"

python/lsst/ts/standardscripts/maintel/home_both_axes.py
iglesu: Not sure I follow. ignore_m1m3 it's basically removed. So there is no "ignore_m1m3" property.

python/lsst/ts/standardscripts/maintel/home_both_axes.py
tribeiro: I mean the `warn_wait`....

python/lsst/ts/standardscripts/maintel/home_both_axes.py
tribeiro: but we added the `disable_m1m3_force_balance` option....

tribeiro: I think we can remove it.

tribeiro (approved):
looks good! thanks

PR #256, opened 2024-12-17
Tickets/DM-47619: Update implementation of the ignore feature in scripts.

MarcoRocchietti:
Update scripts to use the `disable_checks_for_components` method of the `RemoteGroup` class.

### DIFF
--- a/tests/test_auxtel_disable_ataos_corrections.py
+++ b/tests/test_auxtel_disable_ataos_corrections.py
@@ -24,6 +24,7 @@ import random
 import unittest
 
 from lsst.ts import standardscripts
+from lsst.ts.observatory.control.auxtel.atcs import ATCS, ATCSUsages
 from lsst.ts.observatory.control.mock import ATCSMock
 from lsst.ts.standardscripts.auxtel import DisableATAOSCorrections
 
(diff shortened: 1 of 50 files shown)


tests/test_auxtel_disable_ataos_corrections.py
```diff
@@ -40,24 +41,24 @@ async def basic_make_script(self, index):
         self.script = DisableATAOSCorrections(index=index)
         self.atcs_mock = ATCSMock()
 
-        return (self.script, self.atcs_mock)
+        self.script.atcs = ATCS(
```
tribeiro, line 44: These unit tests were doing an actual test of the execution, see that they are using the `ATCSMock` above. Any reason for changing them? If not, I would rather leave them as they were.

tests/test_auxtel_enable_ataos_corrections.py
```diff
@@ -39,6 +40,14 @@ async def basic_make_script(self, index):
         self.script = EnableATAOSCorrections(index=index)
         self.atcs_mock = ATCSMock()
 
+        self.script.atcs = ATCS(
```
tribeiro, line 43: Same as above, any reason for changing the unit tests?

tests/test_auxtel_enable_atcs.py
```diff
@@ -38,9 +38,19 @@ class TestEnableATTCS(
     async def basic_make_script(self, index):
         self.script = EnableATTCS(index=index)
         self.atcs_mock = ATCSMock()
+        self.script._attcs.disable_checks_for_components = unittest.mock.Mock()
```
tribeiro, line 41: Same comment as above, any reason to change the unit test? they should work as is no?

tests/test_auxtel_standby_atcs.py
```diff
@@ -38,9 +38,19 @@ class TestStandbyATCS(
     async def basic_make_script(self, index):
         self.script = StandbyATCS(index=index)
         self.atcs_mock = ATCSMock()
+        self.script._atcs.disable_checks_for_components = unittest.mock.Mock()
```
tribeiro, line 41: Same comment as above. This unit test is using `ATCSMock` any reason for changing it?

tests/test_maintel_mtmount_unpark_mount.py
```diff
@@ -47,18 +47,13 @@ async def test_executable(self):
         await self.check_executable(script_path)
 
     async def test_configure_ignore(self):
-        async with self.make_script():
+        async with self.make_dry_script():
```
tribeiro, line 50: Is this really necessary?

tribeiro (changes requested):
The changes in the scripts and most of the changes in the unit tests looks fine. However, some scripts were testing the ignore feature without any mocking, which I would expect to work with the updated version. 

I made comments in the unit tests that I found, but I am not sure I caught all of them. Can you please, rollback the changes you've made on those? Unless there is really a reason for the changes, in which case, we should probably investigate what the problem is.

tribeiro (approved):
Looks good!

PR #258, opened 2025-01-15
Tickets/dm 47293

iglesu:
This branch represents the core ts_standardscripts functionality


tribeiro (commented):
For this to work, you will have to keep the scripts executable here. So, all the scripts you removed from `python/lsst/ts/standardscripts/data/scripts/`, needs to be recovered and updated to import the script from the new packages.

For the unit tests. I think we will keep the `test_executable` for all the scripts. Maybe we can put those all in a single test file.

doc/news/DM-47293.feature.rst
```diff
@@ -0,0 +1,4 @@
+`ts_standard_scripts` updated to retain only core
```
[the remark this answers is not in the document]
tribeiro: Ping

python/lsst/ts/standardscripts/utils.py
```diff
@@ -46,7 +47,7 @@
 )
 
 
-def get_scripts_dir():
+def get_scripts_dir(file_path: Optional[str] = None):
```
tribeiro, line 50: Is this necessary?

iglesu: good catch, that's no longer necessary

tribeiro:
Hi @iglesu , can you add `lsst-ts/ts_maintel_standardscripts` and `lsst-ts/ts_auxtel_standardscripts` to the list of extra package [here](https://github.com/lsst-ts/ts_standardscripts/blob/4d8c3863f3c3662b728ec241fbe827199f81be65/Jenkinsfile#L5)?

This should allow the unit tests to pass.

tribeiro (approved):
Alright! looks good!

PR #261, opened 2025-02-23
DM-46005 Add AT venting start script.

bbrondel:
This adds a SAL script to start the AT dome venting system. The script works as follows:

1. The script is specifically for the start of venting. It could also be used to stop venting, but I intend to have a second script for that purpose.
2. The script takes two arguments: fan_frequency and gates_to_open.
3. The fan_frequency is the desired fan frequency in Hz. If omitted, no commands are sent to the fan. If zero, the fan is stopped.
4. gates_to_open is a list of the gates that should be opened. Any gates not in this list are commanded to close. If the list is omitted, no commands are sent to the gates.
5. The script should be called when the CSC is ENABLED.
6. The telemetry/events are checked to verify that the commands worked as expected.

### DIFF
--- a/doc/news/DM-46005.feature.rst
+++ b/doc/news/DM-46005.feature.rst
@@ -1,0 +1,1 @@
+Added new script ``atvent_start.py``.
--- a/python/lsst/ts/standardscripts/data/scripts/auxtel/atvent_start.py
+++ b/python/lsst/ts/standardscripts/data/scripts/auxtel/atvent_start.py
@@ -1,0 +1,28 @@
+#!/usr/bin/env python
+# This file is part of ts_standardscripts
+#
+# Developed for the LSST Telescope and Site Systems.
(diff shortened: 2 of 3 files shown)


iglesu:
I split the branch for this PR in two. The original code was done prior to splitting the ts_standardscript in three branches: ts_standardscripts, ts_auxtel_standardscripts and ts_maintel_standardscripts.

So I took @bbrondel work and moved the auxtel part of the code to ts_auxtel_standardscripts and while the execution logic stayed here, in ts_standardscripts.

Brian's original branch was renamed to `tickets/DM-46005-original`

edennihy:
@bbrondel and I discussed that these features were added directly to the prepare_for/vent.py script. This script is no longer needed, and the accompanying PR in ts_auxtel_standardscripts was already closed. I will close this PR.

PR #268, opened 2025-03-27
In base_track_target_and_take_image.py, add option to pass note to co…

edennihy:
…nfiguration.

### DIFF
--- a/python/lsst/ts/standardscripts/base_track_target_and_take_image.py
+++ b/python/lsst/ts/standardscripts/base_track_target_and_take_image.py
@@ -145,6 +145,9 @@ properties:
       - type: string
       - type: "null"
     default: null
+  note:
+    description: A descriptive note about the image being taken.
+    type: string
   camera_playlist:
     description: >-
       Optional name a camera playlist to load before running the script.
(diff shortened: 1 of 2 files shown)


python/lsst/ts/standardscripts/base_track_target_and_take_image.py
```diff
@@ -180,6 +183,7 @@ async def configure(self, config):
         self.config.az_wrap_strategy = getattr(
             self.tcs.WrapStrategy, self.config.az_wrap_strategy
         )
+        self.note = getattr(self.config, "note", None)
```
tribeiro, line 186: Really minor comment for you to consider.

We keep the configuration in `self.config` already so, in principle, you don't need to add a `self.note`. However, if you want to make it easier to reach this parameter, consider making this a property;

```
@property
def note(self):
    return getattr(self.config, "note", None) if self.config is not None else None
```

tribeiro (approved):
Looks good, I left a comment for you to consider though.

python/lsst/ts/standardscripts/base_track_target_and_take_image.py
edennihy: Adopted.

PR #273, opened 2025-04-11
Tickets/dm 46469

iglesu:
Prepare BaseBlockScript for deprecation

### DIFF
--- a/python/lsst/ts/standardscripts/base_block_script.py
+++ b/python/lsst/ts/standardscripts/base_block_script.py
@@ -23,17 +23,10 @@ __all__ = ["BaseBlockScript"]
 
 import abc
 import contextlib
-import hashlib
-import io
-import json
-import os
-import re
-from datetime import datetime
+import warnings
 
 import yaml
 from lsst.ts import salobj, utils
-
-from .utils import get_s3_bucket
 
 IMAGE_SERVER_URL = dict(
     tucson="http://comcam-mcm.tu.lsst.org",
@@ -43,13 +36,27 @@ IMAGE_SERVER_URL = dict(
 
 
 class BaseBlockScript(salobj.BaseScript, metaclass=abc.ABCMeta):
-    """Extend BaseScript to add support for executing blocks.
+    """(Deprecated) Extend BaseScript to add support for executing blocks.
 
     This base class adds a default configuration with reason and program that
     can be provided when executing the script.
+
+    Deprecated:
+        This class is deprecated. BaseScript now supports block metadata
+        directly.
+        Future development should use BaseScript and its built-in block
+        support.
     """
 
     def __init__(self, index: int, descr: str, help: str = "") -> None:
+        warnings.warn(
+            "BaseBlockScript is deprecated. salobj.BaseScript now supports"
+            + " block metadata directly. "
+            + "Future development should use salobj.BaseScript instead.",
+            DeprecationWarning,
+            stacklevel=2,
+        )
+
         super().__init__(index, descr, help)
 
         self.program = None
@@ -125,6 +132,13 @@ class BaseBlockScript(salobj.BaseScript, metaclass=abc.ABCMeta):
         self.program = getattr(config, "program", None)
         self.reason = getattr(config, "reason", None)
         self.test_case = getattr(config, "test_case", None)
+        if self.test_case is not None:
+            warnings.warn(
+                "The test_case functionality in BaseBlockScript is deprecated and"
+                + " will be removed.",
+                DeprecationWarning,
+                stacklevel=2,
+            )
         self.step_counter = (
             utils.index_generator(imin=self.test_case.get("initial_step", 1))
             if self.test_case is not None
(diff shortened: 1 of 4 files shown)


iglesu:
While doing another ticket I just realized I need to test some more cases for this deprecation.

python/lsst/ts/standardscripts/base_block_script.py
```diff
@@ -142,11 +164,22 @@ async def configure(self, config):
     async def get_obs_id(self) -> str | None:
```
edennihy, line 164: As I understand the JIRA ticket, the functionality in this method should be entirely removed. It should just return the deprecation warning and None.

python/lsst/ts/standardscripts/base_block_script.py
```diff
@@ -125,6 +141,12 @@ async def configure(self, config):
         self.program = getattr(config, "program", None)
         self.reason = getattr(config, "reason", None)
         self.test_case = getattr(config, "test_case", None)
+        if self.test_case is not None:
```
edennihy, line 144: The test_case functionality can be removed now, we do not need to leave this for the future.

python/lsst/ts/standardscripts/base_block_script.py
```diff
@@ -217,11 +250,19 @@ async def program_reason(self):
 
     @contextlib.asynccontextmanager
     async def test_case_step(self, comment: str | None = None) -> None:
```
edennihy, line 252: This can be removed along with other test case functionality.

python/lsst/ts/standardscripts/base_block_script.py
```diff
@@ -239,11 +280,20 @@ async def test_case_step(self, comment: str | None = None) -> None:
                 self.step_results.append(step_result)
 
     async def save_test_case(self) -> None:
```
edennihy, line 282: This can be removed along with other test case functionality.

edennihy (changes requested):
A few initial comments for your consideration.

edennihy:
Moving a comment from Igor on JIRA -> @Erik Dennihy The BaseBlockScripts class is used in a bunch of scripts, deprecating it’s methods would also mean moving the below scripts to the salobj BaseScript class, is this within the scope of this ticket?  Alternatively, I thought that setting a warning for now would be a interim solution. Please advice, thanks!

edennihy:
Agreed, the BaseBlockScript class is still in use so we are not yet ready to remove it entirely, but we have migrated some of it's functionality to other parts of the system which is why it is safe to remove it already. 

For example, the `obs_id` that is returned by the `get_obs_id` method is now being handled (and overwritten) by the ScriptQueue here (https://github.com/lsst-ts/ts_scriptqueue/blame/b9a8fb9cbea05d38a6df6393f72d0f66d5de1318/python/lsst/ts/scriptqueue/block_info.py#L105). So, to avoid duplicating that (without fully deprecating the BaseBlockScript yet), we would like that method to return `None` and the deprecation warning for the time being. 

Additionally, the test_case functionality is not used in operations. So we would like to remove that now as well. 

As you note, there are still many scripts that inherit from the BaseBlockScript and some still rely on inheriting the program/reason configuration options that the BaseBlockScript provides, so we're not ready to full deprecate it yet.

iglesu:
> Agreed, the BaseBlockScript class is still in use so we are not yet ready to remove it entirely, but we have migrated some of it's functionality to other parts of the system which is why it is safe to remove it already.
> 
> For example, the `obs_id` that is returned by the `get_obs_id` method is now being handled by the ScriptQueue here (https://github.com/lsst-ts/ts_scriptqueue/blame/b9a8fb9cbea05d38a6df6393f72d0f66d5de1318/python/lsst/ts/scriptqueue/block_info.py#L105). So, to avoid duplicating that (without fully deprecating the BaseBlockScript yet), we would like that method to return `None` and the deprecation warning for the time being.
> 
> Additionally, the test_case functionality is not used in operations. So we would like to remove that now as well.
> 
> As you note, there are still many scripts that inherit from the BaseBlockScript and some still rely on inheriting the program/reason configuration options that the BaseBlockScript provides, so we're not ready to full deprecate it yet.



> https://github.com/lsst-ts/ts_scriptqueue/blame/b9a8fb9cbea05d38a6df6393f72d0f66d5de1318/python/lsst/ts/scriptqueue/block_info.py#L105

Hi @edennihy , I found just one script, that might matter, and uses the test_case_step method, that's in ts_maintel_standardscripts move_p2p.py

https://github.com/lsst-ts/ts_maintel_standardscripts/blob/3093c3ab1c95658a359522ffa62354881bbdf41d/python/lsst/ts/maintel/standardscripts/move_p2p.py#L173

and 
https://github.com/lsst-ts/ts_maintel_standardscripts/blob/3093c3ab1c95658a359522ffa62354881bbdf41d/python/lsst/ts/maintel/standardscripts/move_p2p.py#L191

it seems the test_case_step methods saves the timestamp and the state of each step in an s3 bucket and issues a largeFileObjectAvailable event in the save_case_step method, triggered in the finally statement of BaseBlockScript program_reason method: https://github.com/lsst-ts/ts_standardscripts/blob/9a99bfb117d6cb4e755b839846b8e3af5684f157/python/lsst/ts/standardscripts/base_block_script.py#L216

Could it be that move_p2p script is not use at all? or that the data saved is not used? in that case I could remove that from move_p2p.py, otherwise I could move that script to BaseScript and make sure that that logging info is kept.
Please advice, Thanks!.

edennihy:
Nice detective work! I reached out to the sitcom performance analysis team lead and they confirmed that we have not used the LFA data saved by this script for any of our analyses. So we should be safe there. Maybe you could also just have that method return None for now so that we don't have to update that script until we fully deprecate BaseBlockScript?

iglesu:
> Nice detective work! I reached out to the sitcom performance analysis team lead and they confirmed that we have not used the LFA data saved by this script for any of our analyses. So we should be safe there. Maybe you could also just have that method return None for now so that we don't have to update that script until we fully deprecate BaseBlockScript?

sounds good! let me wrap this up ASAP then, thanks much!

iglesu:
Hi @edennihy latest changes are in, please note that will rebase commits just prior to merging, unless you prefer me to do that before your revision. Thanks!

edennihy (approved):
Approved! Thanks for the updates. Please see my minor comment on the news fragment and remember to squash your commits before merging.

PR #277, opened 2025-05-25
In base_take_image.py, add configurable option to sleep in between im…

edennihy:
…ages.

### DIFF
--- a/doc/news/DM-51095.feature.rst
+++ b/doc/news/DM-51095.feature.rst
@@ -1,0 +1,1 @@
+In ``base_take_image.py``, add configurable option to sleep in between images.


python/lsst/ts/standardscripts/base_take_image.py
```diff
@@ -140,6 +140,10 @@ def get_schema(cls):
                 description: Emulate a slewttime by sleeping before taking data.
                 type: number
                 default: 0
+              sleep_for:
```
dsanmartim, line 143: In the `base_make_calibrations.py` script, we defined a parameter called `wait_between_exposures` for this. I wonder if it wouldn't be better to have something similar here, or maybe `sleep_between_exposures`. It seems more intuitive and better aligned with its own description in L144. Note also that we already use the `sleep_for` keyword in other contexts not related to taking images...

dsanmartim (approved):
It looks good to me. I just added one comment for your consideration.

PR #281, opened 2025-05-30
Tickets/dm 51012

iglesu:
Example of direct link to SAL script documentation:
https://ts-standardscripts.lsst.io/builds/601/available_scripts.html
or via the index file
https://ts-standardscripts.lsst.io/builds/601/index.html

### DIFF
--- a/doc/_static/custom.css
+++ b/doc/_static/custom.css
@@ -1,0 +1,77 @@
+/* Custom CSS for ts_standardscripts documentation */
+
+/* Improve TOC appearance in sidebar - ensure main headers are bold */
+.sphinxsidebar .toctree-l1,
+div.sphinxsidebar ul > li > a,
+div.sphinxsidebar li.current > a,
+div.sphinxsidebarwrapper ul > li > a,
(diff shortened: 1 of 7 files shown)


edennihy:
Following the same logic, could we place the Maintel/Auxtel/OCS general scripts (that are not part of a subfolder) below the scripts with a subfolder? That would match the layout in LOVE.

edennihy:
In a separate topic, when people want to learn about new scripts, the next layer of info after the description is of course the script configuration yaml. I don't think we can add that here since it will be very unwieldy (and it out of scope for the ticket), but instead of providing a hyperlink to the script description in the developer guide could we hyperlink the script itself in github? That might break some rules about documentation linkage, but it would allow users to get the info they need about the available configurations for each script a little faster.

edennihy:
@iglesu in this case I mean a link to the actual yaml. But the more I think about it many scripts inherit from base scripts so a link wouldn't really work. Probably not worth pursuing at this time.

edennihy (approved):

PR #282, opened 2025-06-02
Tickets/dm 50986: Support Summit Observing Weeks 18-19 of 2025

edennihy:


python/lsst/ts/standardscripts/base_track_target.py
```diff
@@ -104,14 +104,20 @@ def get_schema(cls):
                     - dec
                 properties:
                   ra:
-                    description: ICRS right ascension (hour).
+                    description: >-
+                        ICRS right ascension (hour). This coordinate can be passed
+                        as either decimal hours e.g. 12.50 or as a sexagesimal string
+                        e.g "12:30:00.0".
```
dsanmartim, line 110: Just a quick note: the other day, someone asked whether the script accepts coordinates in "hms" and "dms" notation (e.g., 12h30m00.0s -30d30m00.0s). It turns out it does, since base_tcs.py uses Astropy’s Angle class, which deals with the formatting.

I’m mentioning it here in case you want to add it to the description—but part of me thinks it might be better not to, to maintain consistency with the coordinate formats we typically use in almost every block.

dsanmartim (approved):
Approved!

PR #284, opened 2025-06-05
Rename m2 enable/disable_closed_loop

b1quint:


b1quint:
When I rebased this file from `origin/develop`, I got a conflict saying that the two files below were deleted in that branch:
* `tests/test_maintel_m2_disable_closed_loop.py`
* `tests/test_maintel_m2_enable_closed_loop.py`

I removed them here as well. We might want to double-check that.

PR #303, opened 2026-06-17
Added new wait_for_sun_elevation.py script to sleep the queue until s…

edennihy:
…olar elevation is reached.

### DIFF
--- a/tests/test_wait_for_sun_elevation.py
+++ b/tests/test_wait_for_sun_elevation.py
@@ -1,0 +1,83 @@
+# This file is part of ts_standardscripts
+#
+# Developed for the LSST Telescope and Site Systems.
+# This product includes software developed by the LSST Project
+# (https://www.lsst.org).
+# See the COPYRIGHT file at the top-level directory of this distribution
+# for details of code ownership.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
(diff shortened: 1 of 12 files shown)


tribeiro (approved):
looks good!

edennihy:
Tested on BTS on 2026-07-01

python/lsst/ts/standardscripts/__init__.py
```diff
@@ -29,6 +29,7 @@
 from .sleep import *
 from .system_wide_shutdown import *
 from .utils import *
+from .wait_for_sun_elevation import *
```
tribeiro: I am not sure this is necessary... I think we can keep here only things that are broadly used, e.g. the base classes. For now I think we can just stop adding things here for now.

tests/test_wait_for_sun_elevation.py
```diff
@@ -0,0 +1,89 @@
...
+
+from lsst.ts.observatory.control.maintel.mtcs import MTCS, MTCSUsages
+from lsst.ts.standardscripts import BaseScriptTestCase, get_scripts_dir
+from lsst.ts.standardscripts.wait_for_sun_elevation import WaitForSunElevation
+
+
+class TestWaitForSunElevation(BaseScriptTestCase, unittest.IsolatedAsyncioTestCase):
+    async def basic_make_script(self, index):
+        self.script = WaitForSunElevation(index=index)
+        self.script.mtcs = MTCS(
```
tribeiro: I just noticed that `mtcs` is already defined as `DryTest` in the script. I think you can remove this.

